PR #15, opened 2026-07-17
🔖 Publish v0.3.13 — rebuild components against scss! no_hash-capable tairitsu.

langyo:
## Why

All component stylesheets call `scss!` with `no_hash`; tairitsu-macros < 0.5.18 silently ignored the flag and hashed (or, on decimal literals, corrupted) component CSS, so published 0.3.x components render without styling. This re-publishes hikari built against the fixed macro, and fixes build breakage found along the way.

## What

- Workspace version 0.3.12 → 0.3.13; tairitsu floor raised to ^0.5.18 so fresh builds always pick the fixed macro.
- link.scss / typography.scss: `@use 'variables'/'mixins'` pointed at nonexistent neighbours — now `../../../../theme/styles/{variables,mixins}.scss` like the other components.
- radio_group.rs referenced missing `radio_group.scss` (renamed to `radio.scss`) — this alone made hikari-components fail to compile at all.
- build.rs: generate legacy include_str! CSS stubs unconditionally when missing; previously in-workspace local path builds (the documented ~/.cargo/config.toml patch flow) failed with include_str! os error 2.

Known pre-existing limitation (not a regression): theme scss files are not packaged into the published crate, so crates.io builds still degrade the few theme-@use stylesheets; needs a separate packaging solution.

Merge/publish order: tairitsu (0.5.18) → **hikari (0.3.13)** → lagrange (0.2.2).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 target/
 Cargo.lock
 
+# Local-only cargo patches (e.g. path overrides for unreleased tairitsu)
+.cargo/
+
 # E2E test logs
 logs/
 target/e2e_screenshots/
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,6 @@ members = [
     "packages/extra-components",
     "packages/icons",
     "packages/builder",
-    "packages/e2e",
     "packages/i18n",
 ]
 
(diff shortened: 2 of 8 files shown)


langyo:
Closing: requires hikari-components 0.3.13 to be published first, then needs CI fixes.

PR #18, opened 2026-07-17
🐛 Ship packed MDI data so every build gets icon data

langyo:
## Problem

hikari-icons generated its icon set exclusively from `icons/mdi/*.svg` — files downloaded into the workspace at development time and gitignored. Fresh clones therefore built an empty data module, and crates.io consumers got the no-data stub, so `hikari_icons::get()` returned `None` for **every** icon downstream (e.g. lagrange's copy/search/translate icons rendered as nothing).

## Changes

- Pack the 92-icon set into `packages/icons/resources/mdi_icons.dat` (gzip `MDI1`, 7.3 KB) via the existing `scripts/icons/pack_mdi_data.py` and commit it, so the data ships with git and the crates.io package.
- `packages/icons/build.rs` falls back to the packed archive whenever `icons/mdi/` is absent; icon selection keeps its priority (explicit `set` > auto-discovery > whole archive). The empty stub is now only a last resort when the archive itself is missing.
- `hikari-builder` gains `read_packed_icons` + `build_icons_from_packed`, and the SVG module generator falls back to packed records per icon when an SVG is missing.
- Add `MdiIcon::ContentCopy` (downstream renders copy buttons) and dedup alias variants sharing one kebab name in `pack_mdi_data.py`.
- Drop the untracked `packages/e2e` workspace member (fresh clones could not even load the workspace).
- Regression test: core icons resolve to real MDI path data.

## Test plan

- [x] Fresh-clone scenario (no `icons/mdi/`): `cargo test -p hikari-icons` — 92 icons embedded, all tests pass.
- [x] Development scenario (SVGs present): `cargo build -p hikari-icons` — "Icons built successfully", 92 icons.
- [x] `cargo test -p hikari-builder`, `cargo check -p hikari-theme -p hikari-animation -p hikari-palette -p hikari-i18n` all green.
- Note: `cargo check --workspace` still fails on master for `radio_group.scss` / `virtual-scroll.css` — pre-existing, fixed in #15.


langyo:
Closing: requires code fixes to pass CI. Will be handled separately.

PR #61, opened 2026-07-25
✨ Add slot overlay and showTrack prop to HkProgressRing

langyo:
Adds two features to HkProgressRing:\n-  prop (default true) to hide the background track circle when using the ring as an overlay frame\n- Default slot renders content centered inside the ring via  flex-column layout\n- Accepts  alias prop for cleaner API\n\nUsed by shittim-chest PR #79 to display gauge icons + values centred inside progress rings without duplicate background circles.


langyo:
Superseded by #62 (clean branch with correct commit format).

PR #66, opened 2026-07-25
♻️ Rename all component exports to use H prefix.

langyo:
Renames all hikari component exports to use `H` prefix (e.g., `Button` → `HButton`). Runtime/i18n exports and type exports keep their original names.


langyo:
Already applied in master.

PR #74, opened 2026-07-28
📤 Export useReportedTransition from hikari.

langyo:
Exports `useReportedTransition` and its associated types (`ReportedTransition`, `ReportedTransitionTrack`) from hikari's public API (src/index.ts).


langyo:
Already on master via a previous merge.

PR #142, opened 2026-08-09
🎨 Soften the light-theme border from pure black to mid grey.

langyo:
All four light presets used rgb(0,0,0) for the border token, rendering harsh black hairlines around cards, inputs and panels in light mode. Use rgb(120,120,120) so borders read as subtle outlines.


langyo:
Branch re-labelled to langyo author identity; superseded by fix/light-theme-border-langyo.

PR #151, opened 2026-08-10
🎨 Align the vue package specs with the caret-star convention.

langyo:
typescript was pinned as ~6.0 in the dependency spec instead of ^*, which breaks the pnpm frozen-lockfile check in every consumer that links @celestia-island/hikari (e.celestia.world CI: ERR_PNPM_OUTDATED_LOCKFILE). Per §2.4 the pin belongs in pnpm.overrides. Also adds the missing packages field to the workspace yaml (pnpm 9 requires it) and regenerates the lockfile with pnpm 9.15.9 (typescript still resolves to 6.0.3). Verified: frozen-lockfile install passes.

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -54,11 +54,14 @@
     "@vitejs/plugin-vue-jsx": "^*",
     "@vue/server-renderer": "^*",
     "sass": "^*",
-    "typescript": "~6.0",
+    "typescript": "^*",
     "vite": "^*",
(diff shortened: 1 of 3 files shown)


langyo:
Merged with §6.2 note: lint-commits and hosted checks pass; self-hosted Format/Test (nightly) failures are pre-existing runner environment issues (verified: local cargo fmt --check exit 0, changes are frontend package/lockfile only).

PR #152, opened 2026-08-11
🐛 Drive the input ripple animation by timestamp delta instead of frame count.

langyo:
## Problem

On mobile (esp. low-refresh/power-saver mode, ~30Hz rAF), the HkPasswordInput ripple animation appears noticeably slow.

## Root cause

The ripple loop in `packages/vue/src/components/HkPasswordInput.tsx` advanced `rp.radius += R_SPEED` and `rp.peak -= PEAK_SPEED` **per requestAnimationFrame callback** (frame counting), so speed is tied to frame rate: fewer frames per second = slower animation.

## Fix

Drive the animation by timestamp delta:

- Convert `R_SPEED` (0.22/frame) and `PEAK_SPEED` (0.02/frame) to per-second rates (13.2/s, 1.2/s — identical at 60Hz).
- `loop(ts)` records `lastTime`, computes `dt = min(0.1, (ts - lastTime) / 1000)`, passes it to `draw(dt)`.
- `rp.radius += R_SPEED * dt`, `rp.peak = max(0, rp.peak - PEAK_SPEED * dt)`.
- First frame seeded with `performance.now()`; dt clamped to 0.1s to avoid jumps after tab inactivity.

Speed is now constant across refresh rates. Verified with `vue-tsc --noEmit` (only pre-existing vitest module errors on master too).

### DIFF
--- a/packages/vue/src/components/HkPasswordInput.tsx
+++ b/packages/vue/src/components/HkPasswordInput.tsx
@@ -91,10 +91,10 @@ export default defineComponent({
     const GAP = 8;
     const DOT_R = 2.2;
     const SIGMA = 2.8;
-    const R_SPEED = 0.22;
+    const R_SPEED = 13.2;
     const R_WIDTH = 1.3;
     const R_BOOST = 0.7;
-    const PEAK_SPEED = 0.02;
+    const PEAK_SPEED = 1.2;
 
     let COLS = 11;
     let dists: number[][] = [];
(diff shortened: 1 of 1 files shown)


langyo:
Merging with `lint-commits` still pending (>10 min in self-hosted runner queue backlog — environmental, §6.2/CI policy). Local gates all passed: `validate` ✅, hosted fmt+clippy+test ✅, `vue-tsc --noEmit` clean (only pre-existing vitest module errors, identical on master), and the org `celestia-devtools commit-msg-lint` passes locally on both the commit subject and the squash subject (exit 0).

PR #157, opened 2026-08-12
🐛 Restore the password input border, cap its animations at 0.3s and localize its labels.

langyo:
## What

- **Border**: `.hk-pwd-box` now uses the same grey border as HkInput (`1px solid color-mix(in srgb, rgb(var(--color-text)) 14%, transparent))` instead of `border: none`.
- **Animation**: the ripple was driven by `R_SPEED = 13.2` grid-cell-units/s — PR #152's per-frame → per-second change was numerically equivalent (0.22×60 = 13.2), so nothing actually sped up, and the grid scales with box width (a ~300px login box spans ~31 cells → ring needed >1.2s). Now time-driven: the ring sweeps the full grid and the peak decays in a fixed 0.3s budget, independent of box size.
- **i18n**: HkPasswordInput hardcoded English defaults (Password entered / All selected / Caps Lock / Full-width) now resolve through hikari's i18n (`useI18n` + `hikari::passwordInput.*` keys) with optional prop overrides; strength labels (Weak/Fair/Strong) are now translated too. Added `hikari::input.hidePassword/showPassword` for HkInput's toggle.
- **Audit fixes**: HkSearchInput placeholder, HkKeywordSearchModal (title/placeholder/clear/searching/no-matches), HkModal Auto label now go through t().

Locale files updated for all 11 locales (ar/de/en/es/fr/ja/ko/pt/ru/zh-Hans/zh-Hant), JSON + TOML.

## Verification

- `vue-tsc --noEmit` passes (only pre-existing vitest test-file errors, identical to master baseline).
- Chest/arona consume hikari via workspace link, so no changes needed on their side.


langyo:
CI note: the required `lint-commits` check failed at checkout — the self-hosted runner could not fetch the repo (`gnutls_handshake() failed` via the daemon HTTP proxy, same proxy outage that affects node-1; it never reached the lint step). The commit message itself is format-compliant (`🐛 <one English sentence.>`, no colon prefix). `vue-tsc --noEmit` passes locally with no new errors (only pre-existing vitest test-file errors, identical to master baseline). Merging per the documented environmental-failure exemption.

PR #160, opened 2026-08-13
🐛 Keep the entered password on refocus and clear it on first input.

langyo:
## What

The password input wiped the entered password the moment the field regained focus, which was too aggressive. This PR changes the semantics:

- On refocus with an existing value the password is **kept**; the placeholder shows the new localized message `hikari::passwordInput.focusedHasValuePlaceholder` (已聚焦，继续输入将清空原密码 / "Focused, typing will clear the current password").
- The old password is cleared only on the **first edit**: typing/pasting starts from an empty field (`beforeinput` handler clears first), delete/backspace clears the whole field, with an `onInput`/`onCompositionEnd` fallback for engines without `beforeinput`.
- Removed the glow (`drop-shadow`) on the left lock icon in focus and reveal states; the placeholder text breathe animation is kept.

Also adds the first component test for `HkPasswordInput` covering the refocus flows.

Local verification: `vitest run` 46/46, `vue-tsc --noEmit` clean.

Not merging yet — pending user verification.

### DIFF
--- a/packages/vue/src/components/HkPasswordInput.scss
+++ b/packages/vue/src/components/HkPasswordInput.scss
@@ -83,9 +83,7 @@
   display: flex;
   align-items: center;
   justify-content: center;
-  transition:
-    color 0.3s ease,
-    filter 0.3s ease;
+  transition: color 0.3s ease;
 }
 
 .hk-pwd-lock-empty {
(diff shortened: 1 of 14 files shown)


langyo:
CI notes: validate passes; lint-commits fails on the runner environment (pip installs celestia-devtools but the CLI is not on PATH — same documented env failure as recent PRs), not on the commit messages. Local verification: vitest 46/46, vue-tsc 0 errors, and live browser verification on node-2 (login page: refocus keeps value with localized placeholder, clear-on-type, unified 14px italic hints with breathe animation, blur hint = 已输入，点击聚焦将清空已有密码).

PR #169, opened 2026-08-15
🔧 Add the missing node types importer to the hikari vue lockfile.

langyo:
packages/vue/package.json declares @types/node as a dev dependency, but packages/vue/pnpm-lock.yaml had no importer entry, so frozen installs and sibling builds fail with TS2688. Regenerated with pnpm 11.18.0.


langyo:
Hosted fmt+clippy+test PASS, lint-commits and validate PASS. Self-hosted failures are the usual environment/mirror issues; change is lockfile importer only (verified with pnpm 11.18.0 lockfile-only). Merging per AGENTS.md §6.2/§7.

PR #170, opened 2026-08-15
🔧 Allow the hikari npm release to publish with a dirty working tree.

langyo:
The workflow bumps package.json before pnpm publish, which then rejects the unclean tree with ERR_PNPM_GIT_UNCLEAN. Add --no-git-checks.


langyo:
Hosted verification and lint gates PASS. Self-hosted failures are environment-only; workflow-only change. Merging per AGENTS.md §6.2/§7.

PR #171, opened 2026-08-15
🔧 Publish the hikari vue package with npm instead of the OIDC pnpm path.

langyo:
pnpm publish tried OIDC token exchange and failed with 404 before falling back; npm publish uses NODE_AUTH_TOKEN directly.


langyo:
Hosted verification PASS. Self-hosted failure is environment-only; workflow-only change.

PR #172, opened 2026-08-15
🔧 Publish hikari through npm trusted publishing without a token override.

langyo:
The NODE_AUTH_TOKEN override may be shadowing the GitHub OIDC trusted publisher with an invalid repo token. Remove the env override so npm can use trusted publishing.


langyo:
Workflow-only change removing the token override; hosted verification PASS. Self-hosted failures are environment-only. Merging.

PR #173, opened 2026-08-15
🔧 Use node 24 for the hikari npm trusted publishing flow.

langyo:
npm trusted publishing requires npm CLI 11.5.1+ and Node 22.14+; Node 24 is the supported host for the npm 11 CLI. Align setup-node.


langyo:
Workflow-only Node/npm version alignment for OIDC; hosted verification PASS. Self-hosted lint failure is environment-only. Merging.

PR #174, opened 2026-08-15
✨ Scroll overflowing input placeholders like a storefront sign.

langyo:
## Summary
Input placeholders had no overflow handling — text longer than the input line simply clipped. This PR adds two overflow strategies to `HkInput` (+ the shared `HkPlaceholderMarquee` layer):

- `placeholder-variant="marquee"` (default): the placeholder is rendered **three times** in a strip inside a clipping window and translated leftward through the hikari **animation bus** (`onFrame`, unified frame scheduling) — the classic scrolling storefront LED sign, with a seamless wrap because the strip loops modulo one copy. The strip parks when the input is focused (short hold before resuming) or holds a value, and under `prefers-reduced-motion`.
- `placeholder-variant="truncate"`: hard-cut at the container edge with an ellipsis (native placeholder kept).

The overlay only mounts when needed (non-empty placeholder, empty value, enabled); native placeholder attribute is cleared in marquee mode so the two never double-render. Textarea path treated identically. Exported `HPlaceholderMarquee` + `PlaceholderVariant` for reuse.

## Tests
- 9 new tests: 3-copy seam, truncate single copy, empty text, exposed `setActive` API, text-change re-render, HkInput truncate/native-placeholder interplay, default overlay mount, value-gated unmount, disabled guard.
- Full suite 66/66, vue-tsc clean, vite build green.

PR #175, opened 2026-08-15
🐛 Drive the placeholder marquee from the real frame delta.

langyo:
Follow-up to #174 (post-merge review findings).

## Fixes
- **Speed correctness (major)**: the frame callback hardcoded `16ms` per tick, but the animation bus throttles normal-priority entries to a 33ms budget — effective scroll speed was ~48% of the declared `speed` and drifted with frame rate. Now uses `ctx.delta` (seconds) from the bus `FrameContext`, so nominal speed holds at any budget or refresh rate.
- **Test TS18047**: nullable `app._instance` access in the test helper now typechecks clean.
- **Alignment (minor)**: marquee overlay gets `text-align: center` to match the native placeholder posture (HkInput's element is centered) — no horizontal shift between variants.
- **Dead `gap` prop removed**: overflow detection now measures against a `COPY_SPACING` constant kept in sync with the scss copy padding, instead of a prop that never reached the CSS.

## Verification
vue-tsc 0 errors, vitest 66/66, gates unchanged.

### DIFF
--- a/packages/vue/src/components/HkPlaceholderMarquee.scss
+++ b/packages/vue/src/components/HkPlaceholderMarquee.scss
@@ -22,6 +22,9 @@
   line-height: 1.5;
   color: rgb(var(--color-muted));
   white-space: nowrap;
+  // Match the native placeholder's centered posture (HkInput's element is
+  // text-align: center) so the two modes do not shift horizontally.
+  text-align: center;
   // The strip is wider than the window; the JS offset drives translateX.
   will-change: transform;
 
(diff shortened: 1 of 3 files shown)


langyo:
**Environment-only CI failures (exempt per §7 / AGENTS.md):** self-hosted jobs failed with (a) `upload-pack: not our ref` / `ambiguous HEAD` — bare-mirror sync lag for fresh refs (§9.4, rerun of lint-commits now PASS), and (b) the ongoing GitHub-side pack-serving fault for this repo documented on #174 (`remote: bad tree object`, varying shas, all readable via API; hosted fallback green on the same head). Local gates: vue-tsc 0 errors, vitest 66/66; independent review verdict SHIP.

PR #177, opened 2026-08-15
✨ Give the password field the same placeholder overflow strategies.

langyo:
#174/#175 gave HkInput the placeholder overflow layer (marquee / truncate). HkPasswordInput renders its own custom centered placeholder (focus states, breathe animation) with **no overflow handling** — long text was hard-clipped by the box.

This PR:
- Adds `placeholder-variant` prop (default `truncate`): the centered custom placeholder now ellipsizes cleanly by default.
- `marquee` variant layers the shared HkPlaceholderMarquee clipping window inside the placeholder element (same animation-bus-driven scrolling as HkInput).
- Chains the marquee styles.

## Verification
vue-tsc clean, vitest 66/66, version 0.4.6.

PR #178, opened 2026-08-15
🐛 Move the marquee style import to the top of the password stylesheet.

langyo:
#177 inserted the `@use "./HkPlaceholderMarquee"` statement mid-file; Sass requires @use before any rules, so any downstream vite build consuming hikari source failed with a stylesheet parse error. Move it to line 1.

Verified: chest webui vite build now succeeds against this hikari.


langyo:
Hotfix：#177 的 @use 位置违反 Sass 规则，破坏下游源码构建（chest vite build 实证失败→修复后实证成功）。lint-commits PASS、hosted 绿；其余自托管失败为已知镜像延迟/GitHub pack 故障（#174 记录在案）。

PR #179, opened 2026-08-15
✨ Add safe centering overflow sensing and scrollable tabs.

langyo:
## Summary

- **HkScrollContainer `align="center"`** — safe centering: content that fits the viewport is centered via an auto-margin aligner wrapper; overflowing content falls back to the inline-start edge and stays fully reachable by scrolling (unlike `justify-content: center`, which clips the overflowing start). Horizontal axis only; runtime prop toggles supported.
- **Overflow sensing** — the container senses per-axis overflow (`none/start/end/both`) and mirrors it onto the root as `data-h-overflow` / `data-v-overflow` with change-only writes, so consumers can build pure-CSS edge affordances.
- **Optional `fade`** — CSS mask fades only the viewport edges that still hide content, driven by the sensed overflow state; `--hk-scroll-fade-size` is overridable from the root.
- **Exposed API** — `refresh()` (manual re-sense for invisible mutations) and `getOverflow()` join the exposed surface.
- **HkTabs `scrollable`** — wraps the tab list in a centered horizontal HkScrollContainer: centered while it fits, swipe/scroll when it overflows, edge fades hint at hidden tabs, active tab nudged into view on change and on container resize. Consumed next by shittim-chest P25 (mobile status bar + top/bottom tab strips).
- **HkHoverRevealAction** — the reveal chain can now shrink inside width-constrained flex parents (`min-width: 0; max-width: 100%`).
- Version 0.4.5 → **0.4.7** (0.4.6 was taken by #177 on master; rebase kept the bump monotonic).

## Verification

- 3-round subagent verify cycle: R1 PASS (2 should-fix applied), R2 PASS (fixes deep-reviewed), R3 PASS (release readiness). Local: `vitest run` 69/69 green, `vue-tsc --noEmit` clean — re-confirmed after the 0.4.7 re-bump and after rebasing onto master 6f57a98.
- New tests: HkScrollContainer.test.ts (9: aligner rendering, axis gating, fade attr, all four overflow branches via stubbed geometry), HkTabs.test.ts (3: bare list unchanged, scroller wiring, indicator stability).
- Regression audit: default-props DOM delta vs master is exactly two inert `data-*-overflow` attributes; non-scrollable HkTabs markup byte-identical; only internal HScrollContainer consumers (HkPopupSelect, HkErrorBoundary) are vertical/default-align and unaffected; no selector collisions across hikari/chest.
- Noted nits (non-blocking): `scrollable` is mount-static; `fade` is horizontal-only by design.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ hikari currently carries two version lines, plus one stale tag:
   `v0.3.0`–`v0.3.2` and `v0.3.4` were published to crates.io without git
   tags, and `v0.3.15`–`v0.3.19` are unpublished master changes.
 - **Vue port (`@celestia-island/hikari`)** — the `0.4.x` line. The npm
-  package is at `0.4.5` on master but has **never been published** to npm;
+  package is at `0.4.7` on master but has **never been published** to npm;
   publication is pending (A3).
 - **Tag `v0.4.0` is a stale tag.** It was created on an early Vue-port
   commit (2026-07-21) before the npm package version line was established,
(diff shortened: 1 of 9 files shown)


langyo:
**CI environment-failure record (per AGENTS.md §6.2 / §7):**

- Self-hosted jobs (Check Formatting / Clippy nightly / Run tests / Test coverage / Test stable+nightly) all failed at the **Checkout Project** step — the documented bare-mirror sync delay for newly-pushed refs (§9.4, `upload-pack: not our ref` class). No code step ran.
- Hosted runner jobs (Format / Test stable / Test nightly) failed at **Set up job** provisioning (org billing hosted-runner limitation, also documented).
- **Code-level gates are green**: `fmt + clippy + test (hosted)` ✅, `lint-commits` ✅, `Verify version consistency` ✅.
- Local verification on node-1: `vitest run` 69/69 ✅, `vue-tsc --noEmit` clean ✅ (re-run after the 0.4.7 re-bump and after rebasing onto 6f57a98).

Merging under the environment-exemption rule.

PR #181, opened 2026-08-15
🐛 Render the placeholder marquee copies only while the text overflows.

langyo:
## Bug (user-reported)

The login screen's username input showed its placeholder **three times** (one near the middle, two more on the left). The marquee variant rendered its three wrap-around copies unconditionally and relied solely on the `overflow: hidden` clipping window — for a short placeholder (e.g. 用户名) in a wide input, nothing was clipped and all three copies sat side by side. Introduced with the marquee feature in #174.

## Fix (reworked per review: no overlay at all while the text fits)

While the placeholder **fits**, the overlay exists only as a `visibility: hidden` measuring probe and the input shows its **native placeholder** — crisp, screen-reader-visible, zero animation. The moment the probe measures real overflow it reports the flip via a new `overflowChange` event; the host then blanks the native placeholder and the visible strip (three seamless copies, animation-bus-driven) takes over. Window grows back past the text → collapse to hidden probe + native placeholder again.

- Measure the **first copy element** directly (`getBoundingClientRect().width`) instead of `strip.scrollWidth / 3`, which stayed wrong whenever the strip holds fewer than three copies. `visibility: hidden` keeps layout, so the probe keeps measuring through `ResizeObserver`.
- `HkInput`: native placeholder suppressed only while overflowing (both input and textarea branches); stale overflow flag reset when the overlay unmounts (value typed / disabled).
- `HkPasswordInput` (opt-in marquee): its static placeholder text hides while the marquee scrolls so the copies do not overprint it.
- Animation math untouched (offset modulo copyWidth; parked when not overflowing or focused).

## Verification

- `vitest` 8 files / **81 tests** green, incl. reworked marquee suite: hidden single-copy probe while fitting; three copies + visible layer under mocked overflow geometry (driving the component's real `measure()`); collapse back on window growth; HkInput keeps the **native placeholder** while fitting and blanks it on overflow.
- `vue-tsc --noEmit` exit 0. Independent cross-verification: CLEAN.

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.4.8",
+  "version": "0.4.9",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library \u2014 production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkInput.tsx
+++ b/packages/vue/src/components/HkInput.tsx
@@ -64,10 +64,28 @@ export default defineComponent({
       measure(): void;
     }>();
     const isEmpty = computed(() => String(props.modelValue ?? "") === "");
+    // Flipped by the marquee overlay once it measures the placeholder
+    // actually overflowing the input line. While false, the native
+    // placeholder does the showing and the overlay stays a hidden probe.
+    const placeholderOverflows = ref(false);
+    watch([isEmpty, () => props.disabled], ([empty, disabled]) => {
+      // The overlay unmounts on these flips — drop the stale overflow flag
+      // so the native placeholder returns the moment the field is cleared.
+      if (!empty || disabled) placeholderOverflows.value = false;
+    });
 
     const forwardFocus = (active: boolean) => {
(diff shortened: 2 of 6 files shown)


langyo:
**Checks exemption record (AGENTS.md §6.2/§7/§9.4):**
- `Check Formatting` / `Format` / `Clippy (nightly|stable)`: all died at the actions/checkout step with `upload-pack: not our ref` / `possible repository corruption` — bare-mirror sync delay (§9.4), never reached any code. This PR touches only Vue/TS/SCSS files (no Rust).
- `lint-commits`: **pass** (one rerun needed — the first attempt hit the runner's `Value cannot be null (Parameter 'path2')` infrastructure glitch).

Local verification on head 97ecabe: `vue-tsc --noEmit` exit 0; `vitest` 8 files / **81 tests** green (reworked marquee suite: hidden probe while fitting, overflow swap drives the real `measure()`, collapse on window growth, native-placeholder restore). Independent cross-verification verdict: CLEAN (7/7), including the visibility-vs-display deadlock analysis and the no-flicker timing reasoning.

PR #183, opened 2026-08-15
✨ Finish the page header shell and scale button heights per size.

langyo:
## Summary

- **HkPageHeader** completed: big title (+ optional icon chip and subtitle) left, right-aligned `actions` slot, optional `dense` variant, scoped SCSS — the shared in-page shell every chest admin sub-page will use (top app bar keeps identity/navigation only).
- **HkButton height fix**: the base `min-height: 2.5rem` applied to ALL sizes, so sm buttons rendered 40px tall in headers/toolbars. Each size variant now owns its min-height (sm 1.75rem, md 2.5rem, lg 2.75rem).

## Verification

- vue-tsc --noEmit clean; vitest 85/85 (4 new HkPageHeader tests)
- Version 0.4.10 → 0.4.11; CHANGELOG updated

Consumed by the upcoming chest admin redesign PR (联动调试 verified there).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,13 @@ Current master, Rust workspace `0.3.19`.
   excludes `[data-active]` everywhere and shows a primary 10% wash with neutral
   text — hue and font-weight never change on hover (active > hover precedence,
   applied to HkNavItem, HkPopover menu items, and the theme toggle as well).
+- Stop HkButton's base min-height from applying to every size: sm buttons were
+  2.5rem (40px) tall regardless of their compact padding — each size variant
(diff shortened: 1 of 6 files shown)


langyo:
Environment-exempt merge: only `lint-commits` is a required check and it passes. The failing Format/Clippy/Test jobs are (a) Rust workflows untouched by this TS/SCSS-only change and (b) running on runners hit by an infra flake (git `upload-pack: not our ref` via proxy path; `Value cannot be null (path2)` at job setup — runner services restarted, direct SHA fetch from node-1 works). Local verification green: vue-tsc clean, vitest 85/85; integration verified against chest PR #314.

PR #185, opened 2026-08-16
✨ Add submit-on-enter intent to the password and text inputs.

langyo:
## Summary

Auth forms across the family (chest/ERP login, register, setup) wrap their fields in a real `<form onSubmit>` but the visible action button is `type="button"` with an `onClick` handler — so pressing Enter in a field never submits anything. This adds an explicit submit intent to the two auth field components:

- **HPasswordInput / HInput** gain a `submitOnEnter?: () => void` prop. When provided, Enter with no modifiers (and not disabled/readonly) invokes it; the existing `keydown` emit is unchanged so consumers can still intercept.
- No behavior change for existing consumers (prop optional, default undefined).

Consumers wire it as `:submit-on-enter="handleSubmit"` on the password field (and optionally the username field). The chest/ERP agents are picking up the consumer side in parallel.

## Verification

- vue-tsc + vite build green; vitest 93/93 green (existing suites).
- No API breakage: prop optional, no emits changed.

### DIFF
--- a/packages/vue/src/components/HkInput.tsx
+++ b/packages/vue/src/components/HkInput.tsx
@@ -20,8 +20,8 @@ export default defineComponent({
     readonly: { type: Boolean, default: false },
     required: { type: Boolean, default: false },
     name: { type: String, default: undefined },
-  /** Submit intent on Enter (no modifiers) — see HkPasswordInput. */
-  submitOnEnter: { type: Function, default: undefined },
+    /** Submit intent on Enter (no modifiers) — see HkPasswordInput. */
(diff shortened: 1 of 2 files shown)


langyo:
CI note: the failing checks (rustup setup errors — component removal / existing-install path conflicts on the self-hosted runner) reproduce on master as well (runs 31923055997/31923055666/31923055656 all fail the same way before this PR). Environment-only per workspace CI policy §7.1. Local verification: vue-tsc green, vite build green, vitest 93/93 green.

PR #191, opened 2026-08-16
🐛 Close the orphaned rule block left in the date picker styles.

langyo:
## What
The squash of #190 left the tail of a deduplicated trigger block stranded at top level in `HkDatePicker.scss` — an unmatched closing brace that breaks the sass compile for any downstream consumer building from source (shittim-chest's worktree build failed on exactly this). The focus-ring and open-state rules are restored as proper selectors on the trigger, and the file balances again.

## Verification
- packages/vue: vue-tsc 0 errors, vitest 117/117, sass compile of the file exits 0, brace-balance asserted.
- Downstream: shittim-chest worktree `vite build` was failing on this file and passes after the fix.

PR #186, opened 2026-08-16
🐛 Center the toast row and settle its entrance without overshoot.

langyo:
## What
- **One-line toast geometry**: `.hk-toast-item` was `align-items: flex-start` while the icon/close boxes are 1.5rem against a ~1.2rem text line — dead space under single-line messages (fat bottom) and the close glyph visibly lower than the text. Now `align-items: center`: all three optical axes align, height symmetric for single- and multi-line bodies.
- **Enter easing**: drops the back-overshoot spring (`cubic-bezier(.34,1.56,.64,1)`) for a plain ease-out settle — same easing family corrected away on the drawer.

## Verification
- packages/vue: vue-tsc 0 errors, vitest 93/93.
- Downstream visual check lands with shittim-chest PR #316 (consumes this via the sibling worktree link; measured in-browser on dev).

PR #193, opened 2026-08-16
💡 Localize the datetime picker and pass through to native controls on mobile.

langyo:
## Spec

Admin pages carry half-done native picker components. This upgrades both date pickers in the Vue package so desktop keeps hikari's own chrome (with hikari's own i18n layer), while touch-sized viewports pass through to the phone's native OS controls — the better interaction on mobile.

### HkDateTimePicker — the half-done one

1. **Locale-derived names instead of hardcoded English.** `MONTH_NAMES` / `MONTH_NAMES_LONG` / `WEEKDAY_NAMES` literal arrays are gone. Month short/long names, weekday labels and the trigger label are now derived via `Intl.DateTimeFormat` from the shared i18n locale state (same approach as `HkDatePicker.tsx`): `locale = computed(() => useI18n().locale)`, `new Intl.DateTimeFormat(locale, { month: "short" })` etc., and the week-start column from `Intl.Locale` `weekInfo` (with the older `getInfo` and the Sunday fallback HkDatePicker already established). Trigger time stays 24-hour (`hourCycle: "h23"`) to match the steppers. A11y labels still route through `t("hk.dateTimePicker.*")`.
2. **Mobile pass-through.** New prop `nativeOnMobile?: boolean` (default **true**). When true and `useBreakpoint().isMobile` (< 768px), renders a native `<input type="datetime-local">` (or `type="date"` when `showTime=false`) instead of the custom popup. Conversion between the `Date` model and the input's local-time wire format is careful: strict `YYYY-MM-DD[THH:mm]` parsing that rejects rollovers (`2026-02-31`), tolerates engine-appended seconds, preserves clock time when `showTime=false` (mirroring day selection in the custom grid), passes `min`/`max` through in wire format, clamps rejected edits by re-syncing the field to the model, and restores the model value when the field is cleared (the model is a required `Date` with no null semantics). Desktop keeps the custom picker untouched.
3. **Fixed the dead popup mode.** `mode="popup"` previously passed its trigger via `slots.trigger` to `HPopover` — but HPopover renders **only** its default slot, so the trigger never appeared and the popover had no anchor (and `placement`/`offset` were never forwarded). The trigger wrap now renders beside HPopover and feeds it `anchorRef`/`placement`/`offset`.

### HkDatePicker — same mobile pass-through

New `nativeOnMobile` prop (default **true**) rendering `<input type="date">` below 768px. The model already speaks the ISO wire format; the handler emits `null` on clear, rejects out-of-bounds/invalid edits and re-syncs the field to the model.

### Tests

- New `HkDateTimePicker.test.ts` (19 tests) following the `HkDatePicker.test.ts` harness conventions (happy-dom, createApp mounts, `setLocale`): renders with a 42-cell grid, weekday/month labels derive from `Intl` and re-derive after `setLocale("zh-Hans")`/`"de"`, Monday-first week column, day selection preserves time-of-day, min/max disable + reject clicks, `markedDays` dots, popup opens from the trigger, and the full native mobile branch (datetime-local/date types, wire-format value + min/max, model updates, clock preservation, rejection re-sync, cleared-field fallback, popup chrome replacement, opt-out via `nativeOnMobile=false`). The mobile seam is `window.innerWidth`, which `useBreakpoint` reads at setup — no composable mock needed.
- `HkDatePicker.test.ts` extended with 5 mobile-branch tests (native input replaces calendar, opt-out, min/max pass-through, edit + clear emissions, rejected-edit re-sync).

Also: CHANGELOG `Unreleased → Added` entry and npm package version bump 0.4.12 → 0.4.13 (version bumps ride the main PR).

## Verification

From `packages/vue` with `COREPACK_HOME=/mnt/work/corepack corepack pnpm@11.18.0 --config.verifyDepsBeforeRun=false` (worktree reuses the main checkout's `node_modules` via symlink per AGENTS.md §9.6 — no install):

- `exec vue-tsc --noEmit` → **0 errors**
- `exec vitest run` → **13 files / 141 tests passed** (baseline 117 + 24 new: 19 in HkDateTimePicker.test.ts, +5 in HkDatePicker.test.ts)
- `run build` → **success**

Cross-verified by an independent verification subagent (code review, git hygiene, all three commands re-run, public API additivity check, native-input edge cases) — all pass. Not merging; leaving for review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,16 @@ Current master, Rust workspace `0.3.19`.
 
 ### Added
 
+- Add mobile native pass-through to both date pickers and localize the
+  datetime picker: HkDateTimePicker now derives month/weekday names and the
+  trigger label from `Intl.DateTimeFormat` on the active locale (week start
+  from `Intl.Locale` weekInfo, Sunday fallback) instead of hardcoded English
+  tables, its previously dead popup mode renders a real trigger anchored to
+  HPopover (`placement`/`offset` finally wired), and both HkDateTimePicker
+  and HkDatePicker gain `nativeOnMobile` (default true) which swaps the
+  custom popup for the OS `<input type="datetime-local">`/`type="date"`
+  control below the 768px breakpoint with Date-to-wire-format conversion,
+  min/max clamping and model re-sync on rejected edits.
 - Add safe centering and overflow sensing to HkScrollContainer (`align="center"`
   wraps the slot in an auto-margin aligner; `data-h-overflow` /
   `data-v-overflow` mirror sensed overflow; optional `fade` masks the edges
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.4.12",
+  "version": "0.4.13",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library \u2014 production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkDatePicker.scss
+++ b/packages/vue/src/components/HkDatePicker.scss
@@ -267,3 +267,40 @@
   padding-top: var(--hi-space-6, 6px);
   border-top: 1px solid color-mix(in srgb, var(--hi-color-border) 10%, transparent);
 }
+
+/* ── Native mobile input ─────────────────────────────────────── */
+
+.hk-dp-native {
+  display: block;
+  width: 100%;
+  min-height: 2.5rem;
+  box-sizing: border-box;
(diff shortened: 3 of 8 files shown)


langyo:
**CI 豁免记录（§7）**：lint-commits 失败为 `path2` setup 环境性崩溃（与 #186-#191 同款，本仓历史性故障）；其余关键 checks——`fmt + clippy + test (hosted)` 与 `validate` 均 **pass**。本地验证：vue-tsc 0、vitest **141/141**（13 文件，含 19 个新选择器测试：Intl 本地化、mobile 原生透传、日期选择、min/max、setLocale 派生、清空回退）、build 成功；3 轮验证周期完成（round-2 改进在验证中落地后按稳定终树复跑全部）。按策略合并。

PR #194, opened 2026-08-17
🎨 Style the locale picker popup as a proper surface submenu.

langyo:
## Summary

`HkLocalePickerPopup` shipped with **zero styles** — the `hk-locale-picker-*` classes had no scss anywhere, so the popup rendered as bare unstyled buttons (read as "native select" in consuming apps; reported from erp.celestia.world, and shittim-chest has the same unstyled render behind its own wrapper).

This adds `HkLocalePickerPopup.scss`: elevated glass surface (blur + border + dropdown shadow), row hover wash, selected row with primary tint + ✓ affordance, flag column, ellipsised labels, reduced-motion support. Wire-in is a one-line scss import in the component.

## Verification

vue-tsc green; vitest 141/141 green (13 files).

### DIFF
--- a/packages/vue/src/components/HkLocalePickerPopup.scss
+++ b/packages/vue/src/components/HkLocalePickerPopup.scss
@@ -1,0 +1,75 @@
+/* ── Locale picker popup (submenu-style listbox) ────────────────────
+ * Anchored flyout listing the supported locales. Renders as a proper
+ * hikari surface menu: elevated panel, hover wash, selected row with a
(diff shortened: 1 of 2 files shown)


langyo:
CI note: code gates green (Clippy stable 14m38s, Run tests, hosted fmt+clippy+test, lint-commits, version check all pass). Failing jobs are the self-hosted rustup family ('Value cannot be null path2' — the checkout/toolchain env issue that reproduces on master). Environment-only per §7.1.

PR #195, opened 2026-08-17
🐛 Keep the password placeholder layers from doubling on Tab focus.

langyo:
## Bug (user report)
Tab from the previous field into a password input: the ORIGINAL placeholder may not clear, leaving TWO placeholders visible at once.

## Root causes (two, stacked)
1. **Marquee overlay text desync** — `HkPlaceholderMarquee`'s text ternary inside `HkPasswordInput` lacked the `focused` branch: on focus the static layer swapped to the focused prompt while the marquee kept rendering the original placeholder. In marquee mode the `marqueeOverflow` flag is measured from the marquee's (original) text, so the static stayed `visibility:hidden` and the original never cleared on Tab; when the visibility handshake ran a frame behind the re-measure, both texts showed simultaneously.
2. **Swap-patch across states** — the placeholder span persisted across the idle→focused text swap; any mid-flight render could show the stale and fresh text together.

## Fix
- Marquee text now resolves the SAME ternary as the static layer (focused branch included) — the two layers can never disagree.
- The placeholder span is keyed on the resolved state (`idle | focused | has-value`) so a flip re-creates the node instead of patching across it.
- +2 tests: single-text invariant on empty-field Tab focus; static/marquee text lockstep.

## Verification
- vue-tsc 0 errors · vitest **143/143** (13 files) · build ✓.
- Downstream note (chest): its login field also carried `autocomplete="new-password"` on a LOGIN page — that triggers Chrome's own suggest-strong-password inline overlay (a second placeholder-looking text) exactly on Tab focus; chest master now uses `current-password` there (register/setup keep `new-password`).

### DIFF
--- a/packages/vue/src/components/HkPasswordInput.test.ts
+++ b/packages/vue/src/components/HkPasswordInput.test.ts
@@ -240,3 +240,42 @@ describe("HkPasswordInput variants and icons", () => {
     expect(placeholderText(container)).toBe("Confirm password (custom)");
   });
 });
+
+describe("HkPasswordInput placeholder layers on Tab focus", () => {
+  it("shows exactly one placeholder text on focus of an empty field", async () => {
+    const { container, input } = mountPasswordInput("", {});
+    expect(placeholderText(container)).toBeTruthy();
+    const before = placeholderText(container);
+    input.focus();
+    await nextTick();
+    await nextTick();
+    // Exactly one .hk-pwd-placeholder node, holding ONE of the valid texts.
+    const nodes = container.querySelectorAll(".hk-pwd-placeholder");
+    expect(nodes.length).toBe(1);
+    const txt = nodes[0].textContent ?? "";
+    expect([before, "已聚焦，请输入密码", "Focused, enter your password"]).toContain(txt);
+    // No hint/blur overlays doubled on top.
+    expect(container.querySelectorAll(".hk-pwd-blur-hint").length).toBe(0);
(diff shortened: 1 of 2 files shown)


langyo:
豁免记录（§6.2）：lint-commits 失败为已知环境性——分支 08-22 force-push 后 before-SHA 失效（PLAN 记录同款问题），分支上唯一提交 `🐛 Keep the password placeholder layers from doubling on Tab focus.` 经本地 commit-msg lint 复核合规。validate pass。本地验证通过，予以合并。

langyo:
豁免记录（§6.2）：lint-commits 失败为已知环境问题——force-push 后 workflow 事件快照的 base.sha（8ddc795）失效，报 `fatal: Invalid revision range 8ddc795..fc27dd5`（PLAN 已记录同款 before-SHA 失效）。分支唯一提交 `🐛 Keep the password placeholder layers from doubling on Tab focus.` 已本地复核合规（gitmoji + 英文句号结尾）。本地验证 vue-tsc 0 错、vitest 409/409 全绿。予以合并。

langyo:
豁免记录（§6.2，第二次）：lint-commits 仍死于同一环境问题——连续 force-push 使事件快照 base（fc27dd5）失效，workflow 回退分支报 `fatal: Invalid revision range`；其主 range（0b7b318..f6a2499）有效，仅含合规提交 `🐛 Keep the password placeholder layers from doubling on Tab focus.`。本地验证 vue-tsc 0 错、vitest 409/409。予以合并。

PR #196, opened 2026-08-17
✨ Add a generic cascading menu with desktop cascade and mobile sheet modes.

langyo:
## Summary

One primitive for every "menu with submenus" need — locale picking, theme groups, account actions — instead of per-feature popup components:

**`HkMenu`** (`items: HkMenuItem[]` recursive tree; leaf/branch rows, icons/flags, checked/danger/disabled states):

- **Desktop**: root panel anchored to the trigger; rows carrying children cascade to the **right of the anchor row** (flipping left when the viewport runs out) — traditional application menubar behavior. Hovering a sibling collapses deeper levels (classic menubar semantics).
- **Mobile** (< 768px or `mobileBreakpoint`): every level opens as its own **fullscreen sheet**, stacked; each pushed level pushes a history entry so the system/browser **back gesture closes exactly one level** (Android modal navigation mode). Sheets slide in with a one-shot animation (registered in the central animation registry) and honor reduced-motion.
- `HkLocalePickerPopup` is no longer a bespoke component — it is now a thin config over `HkMenu` (same public API, so chest/erp keep working unchanged), and its dedicated scss is absorbed into HkMenu's.

## Verification

- vue-tsc green; vitest **143/143** including 2 new HkMenu tests (desktop cascade into children + leaf select/close, closed-state renders nothing) via the repo's createApp harness.


langyo:
Local verification green (vue-tsc, vitest 143/143 incl. 2 new HkMenu tests). Browser-verified downstream in erp: desktop submenu cascades right of the trigger row (placed left:249/right of anchor at 1280px), mobile fullscreen sheet opens from the drawer language row and closes via its back button. Self-hosted rustup-family checks (when they run) fail on the same env issue as master.

PR #197, opened 2026-08-17
🐛 Fix the cascading menu sibling switching and mobile back-stack semantics.

langyo:
## Summary

Follow-up to #196 (HkMenu). Browser verification of the merged primitive surfaced five real defects against the intended spec ("submenu cascades right of the trigger row like a traditional app menu; mobile = each level is a modal, Android-style back closes ONE level"):

1. **Sibling switching was broken** — `onItem` appended to the path at the *deepest open level* (it did not know the clicked row's level), so clicking `Group One` while `Group Two`'s submenu was open produced path `["group-2","group-1"]` and the submenu silently stayed on Group Two (browser-reproduced). Rows now pass their level; a click replaces the path at that level.
2. **Submenus anchored to the wrong row** — row refs were keyed by level only, so the cascade positioned against the *last* branch row mounted at that level, not the active one. Refs are now keyed `<level>:<key>` and positioning reads the active path.
3. **Mobile back-stack semantics were absent** — the root sheet pushed no history entry and `onPopState` early-returned on our own states, so browser-back closed the WHOLE menu (or did nothing at depth ≥ 2). Now every level — the root included — owns one history entry; back truncates the stack to the landed depth (one level per gesture, root-back closes the menu); in-app back buttons ride the same popstate path. `restoreHistory` uses a single `go(-n)` traversal (consecutive `back()` calls get coalesced by happy-dom and are racy in browsers) with a suppression counter so its landing popstate cannot be misread as a user gesture.
4. **Mode was frozen** — `mobileMode` was a computed with no reactive deps (cached at first read) and the render read `innerWidth` only between renders. A resize tick now feeds the computed; an open menu live-switches sheet↔panel and normalizes its history chain without closing (the transition's own restore previously closed the menu — fixed and regression-tested).
5. **`position: fixed` inside filtered ancestors** — glass popovers (`backdrop-filter`) become containing blocks for fixed children, so a menu opened inside erp's account popover rendered misplaced (verified: style said 249/146, rect was 274/189). Both surfaces now render through `<Teleport to="body">`.

Plus classic menubar hover semantics: hovering a sibling branch switches its submenu in, hovering a leaf collapses deeper levels; root panels with `bottom-*` placements flip horizontally on overflow.

## Verification

- vitest **149/149** (8 HkMenu tests: cascade+select, sibling click-switch, hover-switch/hover-collapse, N-sheets-per-level, browser-back closes one level, in-app back, leaf-close restores history, live viewport switch).
- vue-tsc clean.
- Browser (vite scratch harness, 1280px): 3-level cascade right (x 0→285→514), submenu flip-left at the right edge, active-row anchoring, hover-switch, no history entries pushed on desktop.
- Browser (390px): 1/2/3 levels = 1/2/3 fullscreen sheets at depth 0/1/2; back closes exactly one level each gesture down to closed; leaf select closes all and releases exactly its own entries.
- Consumed in erp (PR celestia-island/erp.celestia.world#28): locale panel anchors exactly at the Language row's right edge (249 = 243+6, top-aligned) inside the glass popover; mobile drawer language sheet is fullscreen, back gesture closes the sheet but not the drawer; locale switch applies + persists.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,14 @@ Current master, Rust workspace `0.3.19`.
 
 ### Added
 
+- Add a generic cascading menu component `HkMenu` (`items` tree with
+  icons/flags/checked/danger/disabled rows) plus `HkLocalePickerPopup`
+  rebuilt as a thin config over it: desktop anchors the root panel to the
+  trigger and cascades submenu panels to the right of their anchor row
+  (flipping left on viewport overflow), switching or collapsing on sibling
+  hover like a classic menubar; mobile renders one fullscreen sheet per
+  level — the root included — where every level pushes a history entry so
+  the system/browser back gesture closes exactly one level.
 - Add mobile native pass-through to both date pickers and localize the
   datetime picker: HkDateTimePicker now derives month/weekday names and the
   trigger label from `Intl.DateTimeFormat` on the active locale (week start
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.4.13",
+  "version": "0.4.14",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library \u2014 production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkMenu.test.ts
+++ b/packages/vue/src/components/HkMenu.test.ts
@@ -37,6 +37,16 @@ const siblingItems: HkMenuItem[] = [
   },
 ];
 
+interface MountedMenu {
+  container: HTMLElement;
+  unmount: () => void;
+}
+
+/**
+ * Menus render through a Teleport to body, so queries go to the document.
+ * Keep at most one OPEN menu mounted at a time (unmount via the returned
+ * handle) — body queries cannot tell instances apart.
+ */
 function mountMenu(
   openRef = ref(true),
   menuItems: HkMenuItem[] = items,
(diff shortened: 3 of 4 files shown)


langyo:
## 3-round verify cycle (record)

- **R1** (independent subagent): SHIP with findings — 3 MINOR (forward-into-spent-stack re-ownership; depth-counted restore landing on a stale entry after a foreign pushState; desktop panels stranded on same-mode resize) + 5 NIT. 149/149, vue-tsc clean.
- **R2** (improvements, c8b9c9d): restore falls back to de-marking the current entry when our marker is no longer current; popstate on our entry while closed replaces the marker (a closed menu never owns live history); idle foreign popstate with nothing pushed early-returns (no spurious update:open(false)); same-mode desktop resize repositions panels. All gates re-run green.
- **R3** (final gate subagent): **SHIP** — traced all three R2 branches sound (buried markers self-heal via de-mark-on-popstate; closed-branch mutations complete before the sync watch reopens; the idle guard never swallows a legitimate collapse since it requires !open). 149/149, vue-tsc clean, CHANGELOG/version sane.
- NIT cleanup (6839f3c): dead `currentLevel` computed removed, restore comment reworded.

Known-accepted NITs (recorded, non-blocking): fixed h=180 flip estimate vs 320px panels; `left-*` never flips rightward; sheets lack dialog a11y/Escape semantics; R2 hardening branches covered only incidentally (follow-up test welcome).

CI note: Test (nightly) failed at "Set up job" (runner provisioning, environment-only — the branch is TS/TSX-only with no Rust changes). lint-commits + validate pass.

PR #198, opened 2026-08-17
✨ Add a sidebar variant to the menu and theme the select surfaces.

langyo:
## Summary

The selector/menu polish wave (user-reported gaps, consumed by erp and chest):

- **`HkMenu variant="sidebar"`** — an inline vertical nav list (no popover, no history): rows with icon/flag/label/badge/active/disabled/danger and a 4px rhythm; root items with `children` render as collapsible groups; groups containing the `activeKey` row default open (seeded, not virtual — the user's first toggle takes over and can collapse them; sibling toggles are independent); deeper levels indent. `HkMenuItem` gains `badge`. Built for admin sidebars and mobile nav drawers.
- **Themed scrollbars** — `HkSelect` popouts (was native chrome), the `HkPopupSelect` viewport (hidden bar → themed thin bar, beating HkScrollContainer's `!important` hide), and `HkMenu` panels/sheets.
- **Small caption labels** — `HkSelect` / `HkPopupSelect` / `HkInput` labels restyled as muted captions (xs / 600 / 60% text), the "tiny title above each field" convention.
- **Option row rhythm** — 2px gaps in select popouts and menu panels so rounded hover pills never touch.

## Verification

- vitest **152/152** (4 HkMenu sidebar tests incl. the active-group collapse regression), vue-tsc clean.
- Browser scratch (1280px): sidebar renders groups/badges/active rows; the region select opens a 20-option popout with a themed thin scrollbar, 2px row rhythm and the caption label.
- 3-round verify: R1 found a MAJOR (auto-expand virtual-state: active group uncollapsible + sibling flicker) — fixed in R2 with a materialized three-set state machine + regression test; R3 pending.

Downstream: erp.celestia.world (labels + PNavSidebar → HkMenu sidebar variant) and shittim-chest /backend sidebar in follow-up PRs.

PR #199, opened 2026-08-17
🔥 Remove the temporary browser verification scaffolds.

langyo:
The menu-polish PR (#198) accidentally carried two untracked scratch files (a vite dev entry and its HTML shell used for browser verification) because `git add -A` staged them. They are dev-only tooling with no imports from the library — removing them restores a clean tree. No behavior change.


langyo:
Policy note: this is a pure file-deletion hotfix (two untracked scratch files removed, no code touched — library files unchanged from master e442bd5, local vue-tsc + vitest unaffected). The Rust check jobs are queue-pending on the self-hosted runners (>15min, org §6.2 environment class); merging per the local-verify + lint-commits gate.

PR #200, opened 2026-08-17
🎨 Lift the field caption labels to 72 percent text alpha.

langyo:
Follow-up to #198: the 60% caption alpha dropped below comfortable contrast on dark-theme tinted surfaces (erp payment modals on wallpaper). Raised to 72% with a `--hk-field-label-alpha` per-app override. vitest 152/152, vue-tsc clean. Note: HkPasswordInput's legacy .hk-pwd-label still uses the old full-size label style — aligning it to the caption pattern is deferred as a separate consistency task.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,9 @@ Current master, Rust workspace `0.3.19`.
 
 ### Fixed
 
+- Raise select/input caption labels from 60% to 72% text alpha (dark-theme
+  tinted surfaces kept the 60% captions below comfortable contrast);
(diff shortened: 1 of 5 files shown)


langyo:
Merge note (user directive: no CI waiting): SCSS/JSON-only change, no Rust code touched. vitest 152/152 + vue-tsc clean locally. The Rust format/clippy failures are checkout-stage 'not our ref' mirror-sync delays on the freshly pushed ref (§9.4 environment class); lint-commits + validate pass.

PR #206, opened 2026-08-17
✨ Add theme extension token groups with hue-clamped pickers.

langyo:
## Summary

Extends the theme system with **namespaced token groups** so downstream apps can register
domain color vocabularies (e.g. SCADA wire/pipe colors) that participate in dark/light modes,
custom themes, and the color-scheme dialog — with **hue-clamped pickers** so physically
meaningful colors (L1/L2/L3 phases, PE yellow-green) can be restyled but never re-hued.

- New `theme/tokenGroups.ts`: `registerTokenGroup` (idempotent, deep-copied), `getTokenGroups`,
  `resolveGroupTokens(mode, overrides)`, circular hue clamping (`clampHue`/`hueDelta`/`wrapHue`,
  `clampRgbToBands`, `clampToSlot`), `groupTokensToCSSVars` emitting `--<group>-<slot>: "r g b"`.
- `setTokenGroupsReapply(fn)` hook wired by `useTheme` at module load: groups registered **after**
  `initTheme()` re-emit their cssvars via a microtask-coalesced re-apply; `tokenGroupsVersion`
  ref makes the dialog react to late registration.
- `ThemePreset.groups?` / `CustomThemePreset.groups?` + `tokensToCSSVars(tokens, groupVars?)`
  (backward compatible); custom themes persist group overrides in localStorage.
- `HkColorSchemeDialog`: collapsible "Extended colors" section per registered group; per-slot
  clamped pickers; `pairWith` slots share one row; prefill via `initialGroups`; values clamped
  on change and on save; renders nothing when no groups are registered.
- `HkColorPicker`: optional `hueClamp` / `sRange` / `lRange` props; every emit path clamps;
  popover shows the allowed hue band with a position thumb; byte-identical behavior without props.
- i18n: `hikari::theme.extendedColors` in all 11 locales; group/slot labels fall back to
  registry strings when keys are missing. hikari itself registers no domain group (mechanism only).

## Test plan

- `pnpm exec vitest run`: **193 passed / 0 failed** (20 new tokenGroups tests + 1 new
  HkColorSchemeDialog component test).
- `pnpm run build` (`vue-tsc --noEmit`): 0 errors.
- Verification highlights (independent cross-check): hue wrap-around math verified
  numerically (center 350°±20° admits hue 10°, rejects 30°); RGB↔HSL round-trip exact over
  140k samples; late-registration re-apply verified at runtime; legacy custom themes without
  `groups` still load; admin-shell files untouched.

Downstream consumer: shittim-chest P34b (SCADA scene view) registers the `scada` token group.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ hikari currently carries two version lines, plus one stale tag:
   `v0.3.0`–`v0.3.2` and `v0.3.4` were published to crates.io without git
   tags, and `v0.3.15`–`v0.3.19` are unpublished master changes.
 - **Vue port (`@celestia-island/hikari`)** — the `0.4.x` line. The npm
-  package is at `0.4.7` on master but has **never been published** to npm;
+  package is at `0.4.18` on master but has **never been published** to npm;
   publication is pending (A3).
 - **Tag `v0.4.0` is a stale tag.** It was created on an early Vue-port
   commit (2026-07-21) before the npm package version line was established,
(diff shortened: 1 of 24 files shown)


langyo:
CI note: `Clippy (nightly)`, `Test (stable)`, `lint-commits` failed at job setup with runner error `Value cannot be null. (Parameter 'path2')` (self-hosted runner provisioning, no steps executed) — environment failure per §6.2. Local gates green (vitest 193/193, vue-tsc 0 errors, 3-round independent verification).

PR #209, opened 2026-08-17
✨ Reveal hover actions on touch and linger for three seconds.

langyo:
✨ Reveal hover actions on touch and linger for three seconds.
  
  HkHoverRevealAction only ever showed its extension on mouse hover, so
  touch devices (mobile status bars, header centres) never got to use the
  revealed action - e.g. the chest bottom bar's add-view plus button.
  
  Touch support:
  - touchstart on the host reveals the extension immediately (tap or swipe
    over a scrollable main slot both work; no flicker while sweeping).
  - touchend / touchcancel schedule a hide after the new touchHideDelay
    prop (default 3000ms), so the action stays visible at least 3s after
    the finger lifts.
  - emulated mouseenter/mouseleave that mobile browsers fire after a touch
    are suppressed during the linger window, so the 140ms mouse delay can
    not undo a touch reveal early.
  
  Mouse behaviour is unchanged (hover reveals, 140ms hide on leave).

### DIFF
--- a/packages/vue/src/components/HkHoverRevealAction.tsx
+++ b/packages/vue/src/components/HkHoverRevealAction.tsx
@@ -9,6 +9,10 @@ export default defineComponent({
   name: "HkHoverRevealAction",
   props: {
     hideDelay: { type: Number, default: 140 },
+    /** How long a touch-revealed extension stays visible after the finger
+     *  lifts. Touch devices have no hover, so a swipe/tap on the host
+     *  reveals the extension and it lingers for this delay before hiding. */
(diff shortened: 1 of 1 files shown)


langyo:
CI 说明：lint-commits 在 self-hosted runner 队列积压（多个并行任务占用 runner），排队超时按 §7 CI 策略处理；提交信息为标准 gitmoji + 单句英文格式，与既往合并一致。本地验证：vue-tsc --noEmit 0 错。hosted fmt+clippy+test 与 validate 已 SUCCESS。变更仅 1 个前端组件文件（触摸支持），不涉 Rust 面。

PR #211, opened 2026-08-18
✨ Add the shared HSignInCard credential form for all front ends.

langyo:
## What

Per the P37 plan (PLAN.md): the sign-in card that every front end hand-rolls moves upstream as a **controlled** composition of the existing auth kit:

- `HkSignInCard` = `HkAuthCard` shell + `HkInput` (username, default user-icon prefix with a slot override) + `HkPasswordInput` (centered placeholder layer, caps-lock / full-width / all-selected hints — all from hikari's own i18n, so consumers pass **zero hint props**) + block `HkAuthSubmitButton`.
- Control contract: fields live inside the card; the consumer injects `@submit(username, password)` and feeds the in-flight state back via `loading`. The card never talks to a backend — erp binds `meLogin`, chest can bind its own auth.
- `hikari::signIn.*` keys (username/password placeholders + submit label) added for all 11 locales (en, zh-Hans, zh-Hant, ja, ko, ru, ar, de, es, fr, pt).
- Logo via `logoSrc` prop or `logo` slot; optional footer slot.
- Version 0.4.19 → 0.4.20 with CHANGELOG.

## Verification

- vitest 212/212 (7 new tests: shell render, submit with username trim, empty-field guard, external loading gate, hard disabled gate, Enter-to-submit from the username field, logoSrc render)
- vue-tsc clean
- erp consumption follows in the erp PR (P37-b)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -178,6 +178,17 @@ Current master, Rust workspace `0.3.20`.
 
 - Remove dead HAuthCard and stale HkStatusBar references. (#88)
 
+## [v0.4.20] - 2026-08-18
+
+### ✨ New
+
+- **HSignInCard** — the shared controlled credential form composing the auth
+  kit (HkAuthCard shell + HkInput with user-icon prefix + HkPasswordInput
+  with the full hint surface from hikari's own i18n + block HkAuthSubmitButton).
+  Fields live inside the card; consumers inject 
+  and feed back .  locale keys ship for all 11
+  languages.
+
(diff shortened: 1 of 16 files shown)


langyo:
Clippy (stable) + Test Coverage failed with `No space left on device` on the self-hosted runner (node-1 root disk at 95% — environmental, not code). Local verification green: vitest 212/212 (7 new SignInCard tests) + vue-tsc clean. Merging per the workspace CI policy (local verify + lint-commits green; environment-only failures recorded).

PR #221, opened 2026-08-18
🛡️ Add touch-action none to the minimap so its drag is never stolen by the surrounding pan surface.

langyo:
Companion to the chest minimap-unification PR: HMinimap is embedded in pannable surfaces (topology overview, media canvas, SCADA scene) whose containers own pan/pinch via `touch-action: none`. The minimap drags itself — declare `touch-action: none` on `.hk-minimap` so mobile browsers never claim its gestures and fire pointercancel mid-drag.

### DIFF
--- a/packages/vue/src/components/HkMinimap.scss
+++ b/packages/vue/src/components/HkMinimap.scss
@@ -11,6 +11,10 @@
   cursor: pointer;
   overflow: hidden;
   pointer-events: auto;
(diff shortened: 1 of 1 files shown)


langyo:
Environment-exemption record: branch pushed via git REST API (github.com web host unreachable from this network; api.github.com works) — CI cannot start for this branch. Merged under the CI-as-reference policy: trivial 4-line scss change, locally verified (vite build in the sibling chest worktree consumes it).

PR #222, opened 2026-08-18
🐛 Center the mobile status-bar tab strip on the true viewport middle.

langyo:
## Summary

User report (mobile): the bottom tab-group buttons were offset to one side instead of centered.

Root cause: the mobile status-bar layout used shrink-flex (left `0 1 auto` / center `1 1 auto` / right `0 0 auto`), which centers the strip inside the leftover space between the wings. Whenever the wings were unequal (compact left connection tag vs right time tray), the strip drifted off-centre.

Fix: both wings now share `flex: 1 1 0` with symmetric justify (start/end), and the center strip is `0 1 auto` at its natural width — the strip stays pinned to the true viewport middle regardless of wing widths, and still hands overflow to the HTabs scroller.

## Verification
- vitest 215/215.
- Desktop layout untouched (its left/right were already symmetric `flex: 1`).

### DIFF
--- a/packages/vue/src/components/HkHoverRevealAction.test.ts
+++ b/packages/vue/src/components/HkHoverRevealAction.test.ts
@@ -1,0 +1,68 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { createApp, h, defineComponent } from "vue";
+
+import HHoverRevealAction from "./HkHoverRevealAction";
+
+/** Mount the component into a detached container and return the host element. */
+async function mountHost(props: Record<string, unknown> = {}) {
+  const host = document.createElement("div");
+  document.body.appendChild(host);
+  const app = createApp({
+    render: () =>
+      h(HHoverRevealAction, props, {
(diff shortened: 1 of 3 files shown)


langyo:
Extending this PR with the third user-reported fix: HHoverRevealAction gains touchRevealMode ('tap' default | 'longpress'). Long-press arms a timer (longPressDelay, 500ms); a quick lift or a slide cancels without revealing — casual tab taps no longer flash the '+' extension. Local: vue-tsc clean, vitest 215/215. Hosted CI failures are the documented org-billing environment class.

langyo:
lint-commits failure is environmental: the self-hosted runner's fetch died with 'gnutls_handshake() failed: The TLS connection was non-properly terminated' (same GitHub-network instability that hit pushes from node-1 today) — not a message-format rejection; the subject follows the gitmoji format and passes local commit-msg-lint. Merging per the local-green + recorded-exemption policy.

PR #223, opened 2026-08-18
🔥 Delete the legacy checkbox stylesheet that pinned the login label vertical.

langyo:
## Summary

User report (both panels): on the new HSignInCard login page the "记住登录 / Remember login" checkbox label rendered vertically (one character per line).

Root cause: the pre-migration legacy stylesheet (`packages/components/src/styles/components/checkbox.scss`, plus an orphaned duplicate at `styles/checkbox.scss`) styles `.hk-checkbox-md { width: 20px; height: 20px }` treating the class as the BOX. The current HkCheckbox component puts `hk-checkbox-md` on its LABEL wrapper (box geometry lives on `.hk-checkbox-box`). With the legacy sheet still bundled, the label got pinned to a 20×20 column → four CJK glyphs stacked vertically.

Verified zero consumers of the legacy class structure (`hk-checkbox-label/checked/disabled/text`, `hk-checkbox-sm/lg` sizing) repo-wide and in downstream chest/erp. The sheet is dead weight + actively harmful; this removes it and its index import.

## Verification
- vite build clean, vitest 219/219.
- Live repro: demo login page label computed 20×20px → with this fix the label returns to inline-flex natural width.


langyo:
Required checks (lint-commits, validate) PASS. The remaining failures are environmental: the Rust toolchain jobs die at 'curl sh.rustup.rs SSL_ERROR_SYSCALL' and mirror-sync lag (fixed by manual mirror fetch) — the documented GitHub-route instability window class (see PLAN env-failure archive); this PR touches zero Rust code (2 scss deletions + 1 import removal). Local: vite build clean, vitest 219/219. Merging per local-green + recorded-exemption policy.

PR #216, opened 2026-08-18
🔥 Remove the unused Vue connection-status chain and restore the Rust ink-bar slider.

langyo:
Follow-up dead-code audit of the component library after #214 (user-directed sweep):

**Audit method**: cross-referenced every legacy stylesheet class against all emitters (Rust rsx + palette enum suffixes — `hk-` + as_suffix(), Vue components, and all downstream repos). Result: **0 dead classes remain** in the legacy stylesheets after #214 — the indicator block was the only one.

**What WAS dead and is removed here**:
- `HConnectionStatus` / `HkStatusBar` / `HkCountdownDigit` component chain — zero consumers anywhere (every front end vendors the plana copies instead); exports, plana-flip animation registrations and parity assertions removed with them
- 11 `connection.json` locale files (`hikari::statusBar.*`) — eagerly bundled into every consumer via the i18n glob, consumed only by the dead chain
- `HBackendStatus` type export (dead with the chain)

**Bug fix riding along**: the Rust Tabs slider (`.hk-tabs-ink-bar`) had NO styling since forever — the original neon rules were authored for `.hk-tabs-indicator`, a class no Rust markup ever emits (naming mismatch). The slider styling now targets the element the component actually renders, restoring the intended neon underline for the lagrange/tairitsu SSR line.

vitest 215/215, vue-tsc clean, sass compile clean. Version 0.4.22 → 0.4.23.

PR #225, opened 2026-08-18
🔥 Remove the unused Vue connection-status chain and restore the Rust ink-bar slider.

langyo:
Repost of #216/#224 (merge gate stuck reporting the required lint as 'expected'; clean single-commit branch off current master).

Dead-code sweep after #214 (user-directed): legacy stylesheet audit found 0 remaining dead classes; removed the truly dead Vue chain HConnectionStatus/HkStatusBar/HkCountdownDigit (zero consumers — every front end vendors the plana copies) plus its 11 connection.json locale files that the i18n glob eagerly bundled into every consumer, and the HBackendStatus type export. Bug fix riding along: the Rust Tabs .hk-tabs-ink-bar slider had no styling since the original neon rules were authored for .hk-tabs-indicator (a class no Rust markup emits) — the styling now targets the real element, restoring the slider for the lagrange/tairitsu SSR line. vitest 215/215, vue-tsc clean, sass clean. Version 0.4.23.


langyo:
Root cause found: reusing head SHA 6e0d30d carried the stale FAILED lint check-run from the #216 era alongside the new SUCCESS one — branch protection sees the failure. Reposted on an amended SHA with a clean check history.

PR #228, opened 2026-08-19
🐛 Reorder the mono font stack so Fira-Code-less devices keep monospace digits.

langyo:
## Problem

`--font-mono` put `var(--font-sans)` directly after "Fira Code" (the intent: CJK glyphs fall back to the UI CJK font, avoiding kai-style system mono CJK). But **Fira Code is a system font — phones never have it installed**, so on every mobile device the latin/digits silently rendered in the proportional sans, breaking every mono alignment in consuming apps (status-bar version hashes, latency readouts, code samples).

## Fix

Sans now sits right **before the generic `monospace` keyword**:

- latin/digits: Fira Code → ui-monospace (SF Mono/Roboto Mono on phones) → JetBrains Mono → …
- CJK glyphs (uncovered by latin mono faces): still fall to `--font-sans` — the original kai-avoidance intent preserved
- generic `monospace` remains the last resort only

Both definitions aligned: `packages/vue/src/styles/admin-tokens.scss` and `packages/theme/styles/_layout.scss`.

## Verification

- `vue-tsc --noEmit` ✅, `vitest run` 215/215 ✅
- Downstream: erp's vendored copy of the same stack fixed in its own PR; chest consumes this via the sibling link

### DIFF
--- a/packages/theme/styles/_layout.scss
+++ b/packages/theme/styles/_layout.scss
@@ -19,7 +19,10 @@
 
   // ── Font families (body text, mono for version/numbers) ─
   --font-sans: "Inter", ui-sans-serif, system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, sans-serif;
-  --font-mono: ui-monospace, "JetBrains Mono", "Fira Code", Menlo, Consolas, monospace;
+  // Mono faces first for latin/digits; sans sits before the generic keyword
+  // only so CJK glyphs (uncovered by latin mono faces) fall back to the UI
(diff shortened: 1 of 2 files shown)


langyo:
**CI exemption record (per PLAN §7 CI 使用策略):** All Rust jobs fail at the `Setup Rust` step — `static.rust-lang.org … tls handshake eof` through the runner proxy (the same rustup TLS window recorded earlier today for erp's check job). Environmental, not code: this PR touches two SCSS font-stack lines only. Local verification green: `vue-tsc --noEmit` ✅ and `vitest run` 215/215 ✅.

PR #229, opened 2026-08-19
🐛 Restore the touch longpress reveal mode that chest's footer consumes.

langyo:
## Problem

#226 dropped `touchRevealMode`/`longPressDelay` from `HkHoverRevealAction` as part of the dead connection-status chain cleanup — but the mode is **not dead**: shittim-chest #355 gates its mobile footer add-view reveals behind a long press, and chest master fails to compile against hikari master (TS2322 `touchRevealMode does not exist` on ChatLayout.tsx:430 and StatusBar.tsx:129). The zero-consumer claim only held for the connection-status chain, not the longpress mode that landed in #222 for chest's #355.

## Fix

Restores the props, the long-press timer wiring, and the four component tests verbatim from 10cb583 (pre-#226).

## Verification

- `vue-tsc --noEmit` ✅
- `vitest run` **219/219** ✅ (215 current + 4 restored longpress tests)

### DIFF
--- a/packages/vue/src/components/HkHoverRevealAction.test.ts
+++ b/packages/vue/src/components/HkHoverRevealAction.test.ts
@@ -1,0 +1,68 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { createApp, h, defineComponent } from "vue";
+
+import HHoverRevealAction from "./HkHoverRevealAction";
+
+/** Mount the component into a detached container and return the host element. */
+async function mountHost(props: Record<string, unknown> = {}) {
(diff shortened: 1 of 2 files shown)


langyo:
**CI exemption record (per PLAN §7 CI 使用策略):** All Rust jobs fail in setup — `Setup Rust` exit 35 (rustup TLS handshake eof via the runner proxy, same window as #228/erp#45) and apt exit 100 (network). Environmental, not code. Local: `vue-tsc --noEmit` ✅, `vitest run` 219/219 ✅ (includes the 4 restored longpress tests).

PR #230, opened 2026-08-19
📱 Dock every mobile modal as a bottom sheet with a reserved top band.

langyo:
## Summary

Two user-reported mobile issues + a systematic unification:

1. **Overlay-over-sheet z bug**: the blurred overlay could paint above the modal content while the popup-manager handle resolves (inline z briefly 0/1). Explicit sibling stacking: overlay z:0, content z:1 — content is structurally always above.
2. **Unified mobile form factor**: every modal on phones (≤767px, matching useBreakpoint().isMobile) docks as a **bottom sheet** — rises from the bottom edge (translateY reveal), top corners rounded only, full width, and never grows past the top band reserved for the secondary-window breadcrumb strip (`--hk-sheet-top-inset`, default 4.25rem; consumers can override). Footer honours the iOS safe-area inset. Replaces the old ≤640px full-screen takeover.

Consumers: chest injects the breadcrumb-reserve value (chest-side PR follows); other repos inherit the default.

## Verification
vue-tsc clean; vitest 219/219.

### DIFF
--- a/packages/vue/src/components/HkModal.scss
+++ b/packages/vue/src/components/HkModal.scss
@@ -23,6 +23,11 @@
 .hk-modal-overlay {
   position: fixed;
   inset: 0;
+  // Explicit stacking inside the root: the content sibling is ALWAYS above
+  // the overlay regardless of the popup-manager base handed to the root
+  // (fixes the blurred overlay painting over the sheet when the manager
+  // handle hasn't resolved yet — zIndex briefly starts at 0).
+  z-index: 0;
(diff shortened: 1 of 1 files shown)


langyo:
Required checks PASS (lint-commits, validate). The optional Rust Lint job dies at rustup download (curl exit 35, SSL_ERROR_SYSCALL — the documented network-window environment class; this PR is SCSS-only, zero Rust). Local: vue-tsc clean, vitest 219/219. Merging per local-green + recorded exemption.

PR #231, opened 2026-08-19
🐛 Drop the UI sans from the mono stack so no-concrete-mono environments stay monospace.

langyo:
## Problem

The v1 reorder in #228 (`var(--font-sans)` before the generic `monospace` keyword) still fails on environments with **no concrete mono face installed** — stock Linux desktops and most Androids (`ui-monospace` is Apple-only): the proportional sans wins before the generic fallback, so digits still render proportionally (pixel-verified on a Linux host: two 14-char hashes measured 87.5px vs 94.34px).

## Fix

Pure mono stack — sans removed entirely, Linux monos added, generic `monospace` as the final catch-all:

- Apple → ui-monospace / SF Mono; Windows → Consolas; Linux → DejaVu/Liberation Sans Mono; Android → generic monospace (Roboto Mono-ish)
- CJK glyphs fall through to the platform last-resort CJK font (= the UI sans CJK on every modern platform — the kai concern the original comment guarded against)

Both definitions aligned (`admin-tokens.scss`, `theme/_layout.scss`).

## Verification

- `vue-tsc --noEmit` ✅, `vitest run` 219/219 ✅
- Pixel check on a Linux host (no Fira Code/JetBrains Mono installed) after this change lands in the consuming app follows in the P44 deploy verification

### DIFF
--- a/packages/theme/styles/_layout.scss
+++ b/packages/theme/styles/_layout.scss
@@ -19,10 +19,12 @@
 
   // ── Font families (body text, mono for version/numbers) ─
   --font-sans: "Inter", ui-sans-serif, system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, sans-serif;
-  // Mono faces first for latin/digits; sans sits before the generic keyword
-  // only so CJK glyphs (uncovered by latin mono faces) fall back to the UI
-  // CJK font instead of the system's kai-style monospace CJK.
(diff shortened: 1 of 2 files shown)


langyo:
**CI exemption record (per PLAN §7 CI 使用策略):** All Rust jobs fail in setup (rustup TLS eof via runner proxy — the same window recorded on #228/#229, erp #44/#45, chest #361/#367 today). Environmental. Local: `vue-tsc --noEmit` ✅, `vitest run` 219/219 ✅. SCSS-only change.

PR #233, opened 2026-08-19
🛡️ Exclude hikari component interiors from the remaining element-level control rules.

langyo:
## Problem

The select chevron baseline, the label block rule, and the button-base element rule were the last bare-element form/control selectors in `theme/styles/base.scss` without the hk- exclusion that #232 added to the input/textarea baseline — flagged as the advisory in #232's verification round.

Same specificity hazard as the double-border input bug: the element-level `button:focus-visible` ring (0,1,1) outranks the component's plain-class styles on every hk- button interior (`hk-btn` / `hk-icon-button` / `hk-tabs-trigger`), and the label block's `margin-bottom` can leak into `.hk-input-label` where the class doesn't reset it. No hikari component renders a bare `<select>` today, so the chevron guard is purely symmetric defense.

## Fix

The same `:not([class^="hk-"]):not([class*=" hk-"])` exclusion applied to all three remaining rules (`select`, `label`, `button`). Bare app markup keeps the baseline; every hikari component interior is untouched.

## Verification

- Selector-chain logic identical to #232's (verified pattern)

### DIFF
--- a/packages/theme/styles/base.scss
+++ b/packages/theme/styles/base.scss
@@ -435,7 +435,11 @@ textarea:not([class^="hk-"]):not([class*=" hk-"]) {
     }
 }
 
-select {
+// Same hk- exclusion as the input/textarea/select baseline above: hikari
+// components render bare <select>/<label>/<button> interiors that own their
+// chrome, and the element-level :focus-visible ring (0,1,1) would outrank
+// their plain-class styles exactly like the double-border input bug (#232).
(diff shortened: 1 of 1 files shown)


langyo:
**CI exemption record (per PLAN §7 CI 使用策略):** All Rust jobs fail in setup — rustup TLS eof via the runner proxy, the same environmental window recorded on #228-#232 today. Local gates green: `vue-tsc --noEmit` ✅, `vitest run` 219/219 ✅. SCSS-only change.

PR #234, opened 2026-08-19
🛡️ Register the select and menu overlays with the popup z context.

langyo:
## Problem

The z hierarchy existed in THREE duplicated token sets that disagreed (`--z-header` 30 vs 100, `--z-sidebar` 40 vs the HkSidebar component default 900), and two overlay components escaped the registered context: **HkSelect's popout hardcoded z 1100** and **HkMenu's panels/sheets 1200+level** — values inside the popup-manager band (1000+2n) but unregistered, so a later-opened modal overtakes them (select-in-modal gets covered).

## Fix

- **Canonical band table** in `theme/styles/_layout.scss` with explicit group bands — local 0-9 / content-float 10-99 / app-chrome 100-199 / app-overlay 200-899 / app-sheet 900-999 / popup-stack 1000-1999 (manager) / toast 9999 / tooltip 10000 / fatal 10001 / titlebar 100001 / breadcrumb 100002. `admin-tokens.scss` mirrors the same values; `.s-status-bar` stops misusing `--z-sidebar` (gets `--z-footer`).
- **HkSelect / HkMenu register** with `usePopupManager` (kind=dropdown, 1000+2n) — their teleported roots take the allocated z inline; the hardcoded SCSS values are gone. HkMenu's mobile sheet levels now stack by DOM order inside the registered context (the per-sheet `` was redundant once the root owns a context).
- **HkLocalePicker**: scoped dropdown (absolute inside the header's own stacking context, not teleported) — inline 1000 → `var(--z-header-popup, 150)`, documenting the band it conceptually belongs to without racing the popup stack.

## Verification

- `vue-tsc --noEmit` ✅ · `vitest run` 219/219 ✅
- Companion PRs align chest/erp's vendored token blocks to the same band table

PR #235, opened 2026-08-19
🛡️ Move the admin shell footer onto the footer band.

langyo:
## Problem

HkAdminShell's footer slot wrapper still carried the pre-band inline `zIndex: 40`, so auth/admin layouts rendered a nested double footer (outer shell at 40, inner PFooter content at 110) after the band wave (#234 + chest #373).

## Fix

The shell footer now uses `var(--z-footer, 110)` — the same band token the vendored footers use.

## Verification

### DIFF
--- a/packages/vue/src/components/HkAdminShell.tsx
+++ b/packages/vue/src/components/HkAdminShell.tsx
@@ -107,7 +107,8 @@ export const HkAdminShell = defineComponent({
         </div>
 
         {slots.footer && (
-          <footer class="s-status-bar" style={{ position: "fixed", bottom: 0, left: 0, right: 0, zIndex: 40 }}>
+          <footer class="s-status-bar" style={{ position: "fixed", bottom: 0, left: 0, right: 0, // Footer band (was a bare 40 — the pre-band chrome value).
+            zIndex: "var(--z-footer, 110)" }}>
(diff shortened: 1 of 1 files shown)


langyo:
**CI exemption record (per PLAN §7 CI 使用策略):** Self-hosted Rust jobs fail at `Setup Rust` — the runner's rustup toolchain install is broken (`failure removing component 'cargo-x86_64-unknown-linux-gnu', directory does not exist: 'share/zsh/site-functions/_cargo'` — same signature as #234's reruns). Environmental. `fmt + clippy + test (hosted)` **PASS** + `validate` PASS. Local equivalents on the same node: `cargo fmt --all -- --check` ✅, `cargo clippy --workspace -- -D warnings` ✅, `cargo test --workspace` (running, results in follow-up), plus `vue-tsc` ✅ and `vitest run` 219/219 ✅ for the touched package.

langyo:
**Follow-up on the exemption:** `cargo test --workspace` was killed locally as overkill for this change — the PR touches one TSX line in packages/vue (an npm-workspace package verified by vitest 219/219 ✅); no Rust crate compiles this file. `cargo fmt` ✅ + `cargo clippy --workspace -- -D warnings` ✅ both ran green on this node.

PR #238, opened 2026-08-19
🎨 Give the gauge ring a palette-context variant API and stop the icon button from filling stroke icons.

langyo:
## What

- **HkGaugeRing** becomes a real theme-system citizen:
  - New `variant` prop (`primary | success | warning | danger | info | muted | auto`) — colors resolve from the theme palette context via the bridged `--hi-color-*` token family (`styles/admin-tokens.scss` bridges app `--color-*` triplets onto it; hikari-standalone gets foundation defaults). `auto` colors by utilisation thresholds (>= 90 danger, >= 75 warning, else success).
  - New single-ring `value` prop alongside the multi-ring `rings` API; explicit `color`/`trackColor` overrides still win (inline style).
  - Self-contained center overlay: `.hk-gauge-ring__center/__value/__label` replace the Tailwind-style utility classes (`absolute inset-0 flex … text-lg`) that only existed in hosts with an atomic-CSS layer — in any other host the center content drifted beside the ring at default font sizes. Typography now scales with `size` (value = clamp(11, size/6, 22)px), tabular figures for live ticks.
  - Aria: single-value rings expose `progressbar` semantics with `aria-valuenow`; multi-ring gauges are `img`.
- **HkIconButton** fixes:
  - Stop forcing `fill: currentColor` on slotted SVGs — CSS beats the SVG presentation attribute, so every lucide (stroke-based) icon rendered as a filled blob (the workspace-manager refresh glyph was the visible casualty).
  - `inheritAttrs: false` silently dropped caller attrs: `aria-label` never reached the DOM (icon-only buttons had no accessible name) and caller `class` was discarded. Attrs are now spread and the class merged.
  - 24px size class icon token xs(12px) → sm(14px): a 12px glyph in a 24px hit target read as a speck.

## Verification

- vitest 229/229 (incl. 10 new: variant mapping, auto thresholds, clamp, aria roles, center classes, icon-button attr fallthrough/class merge/click/slot)
- vue-tsc --noEmit clean

Chest follow-up consumes the variant API in the bridge-node manager and the admin system info page.

### DIFF
--- a/packages/vue/src/DemoApp.vue
+++ b/packages/vue/src/DemoApp.vue
@@ -32,6 +32,11 @@
         <HIconButton icon="search" variant="ghost" :size="16" />
         <HIconButton icon="settings" variant="secondary" :size="24" />
         <HIconButton icon="heart" variant="primary" :size="40" />
+        <!-- The icon slot accepts any component (lucide etc.); attrs like
+             aria-label now fall through to the button element. -->
+        <HIconButton variant="ghost" :size="24" aria-label="Refresh">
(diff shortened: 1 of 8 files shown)


langyo:
Environmental note: the self-hosted jobs (Test stable/nightly, Clippy, Format, Lint, Coverage, Check Formatting) all fail at the `actions/checkout` step — bare-mirror `not our ref` sync delay for the fresh branch, documented class in PLAN §7. The hosted job (`fmt + clippy + test (hosted)`) passes. Local verification on node-1: `vitest run` 229/229 (incl. 10 new tests), `vue-tsc --noEmit` clean.

PR #239, opened 2026-08-19
📝 Merge every class binding form into the icon button and align the icon size ladder doc with the foundation tokens.

langyo:
## What

Sweeps the two leftovers flagged by the P48 cross-verification report:

- **HkIconButton class merge now handles every Vue class-binding form.** The #238 fix merged only string/array forms; an object-form binding (`:class="{ active: cond }"`) was silently dropped. A tiny `classesOf` normalizer (string | array | object, recursive — the same semantics Vue applies to class bindings) feeds the computed class list. No caller is affected today (all chest call sites use string form), this closes the robustness gap.
- **docs/zh-Hans/design/layer1.md icon-size ladder realigned with foundation.scss.** The doc listed xs=12/sm=16/md=20/lg=24/xl=32; the tokens are xs=12/sm=14/md=16/lg=20/xl=24 — only xs matched. The whole ladder is corrected and annotated as mirroring `theme/styles/foundation.scss`. Other locales record no literal values (zh-Hans is the only affected file); layer2/button docs reference `var(--hi-icon-size-*)` symbolically and stay correct.

## Verification

- vitest 230/230 (new case: array-form `["alpha", {beta: true, gamma: false}]` and object-form `{delta: true, epsilon: false}` bindings merge correctly; falsy keys excluded; component classes survive)
- vue-tsc --noEmit clean

### DIFF
--- a/docs/zh-Hans/design/layer1.md
+++ b/docs/zh-Hans/design/layer1.md
@@ -182,12 +182,12 @@ packages/theme/styles/foundation.scss
 
 ```scss
 :root {
-  // 图标尺寸
+  // 图标尺寸（与 theme/styles/foundation.scss 一致）
   --hi-icon-size-xs: 12px;
-  --hi-icon-size-sm: 16px;
-  --hi-icon-size-md: 20px;
-  --hi-icon-size-lg: 24px;
-  --hi-icon-size-xl: 32px;
+  --hi-icon-size-sm: 14px;
+  --hi-icon-size-md: 16px;
+  --hi-icon-size-lg: 20px;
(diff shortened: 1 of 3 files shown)


langyo:
Environmental note (PLAN §7 class): self-hosted jobs fail at `actions/checkout` / toolchain setup (bare-mirror `not our ref` sync delay for the fresh branch — same class as #238). The hosted job (`fmt + clippy + test (hosted)`) passes, as do `lint-commits` and version consistency. Local verification on node-1: vitest 230/230 (incl. the new class-binding form case), vue-tsc --noEmit clean.

PR #242, opened 2026-08-20
🐛 Contain markdown renderer overflow on mobile and rescue unseparated tables.

langyo:
## Problem (reported from dev.celestia.world mobile)
1. Report cards exceeded the phone viewport width — the markdown renderer had no `min-width: 0`/`overflow-wrap` containment, so a single long token (URL, JSON fragment leaked from agent reports) stretched its flex card past the screen.
2. Tables stayed as raw `| --- |` text — (a) agent reports often paste a GFM table directly after a prose line (no blank line), which marked emits as a plain paragraph; (b) even when parsed, a wide table stretched the card instead of scrolling.

## Changes (`HkMarkdownRenderer`)
- Container: `min-width: 0; max-width: 100%; overflow-wrap: break-word; word-break: break-word` (+ body inheritance) — the flex-child shrink fix.
- `rescueTables()` pre-pass: inserts the missing blank line when a table header + delimiter row follows a non-table line, so marked's GFM tokenizer recognizes it.
- Table wrap: post-processes the sanitized HTML to wrap every `<table>` in `.hk-md-table-wrap` (horizontal scroll container, `-webkit-overflow-scrolling: touch`). Post-processing instead of a renderer override because marked 18's table renderer receives structured cell tokens; re-implementing cell rendering would drift from upstream.
- Wide-table CSS: wrapper scrolls; inner table keeps a `min-width` so narrow tables still look like tables.

## Consumers
Every `HMarkdownRenderer` use site benefits automatically — shittim-chest report cards + report detail modal (both render through this component).

## Verification
- `vue-tsc --noEmit` (packages/vue): 0 errors
- Manual table-rescue unit verified against the raw agent-report sample (JSON envelope + table adjacency) from the live incident

### DIFF
--- a/packages/vue/src/components/HkMarkdownRenderer.scss
+++ b/packages/vue/src/components/HkMarkdownRenderer.scss
@@ -119,7 +119,7 @@
 
 .hk-md-table-wrap table {
   margin: 0;
-  min-width: 24rem;
+  min-width: fit-content;
 }
 
 .hk-markdown th,
--- a/packages/vue/src/components/HkMarkdownRenderer.scss
+++ b/packages/vue/src/components/HkMarkdownRenderer.scss
@@ -3,10 +3,20 @@
   line-height: 1.7;
   color: var(--hi-color-text-primary, #c9d1d9);
   position: relative;
+  /* Mobile containment: the renderer often lives inside flex/grid cards.
+     Without min-width:0 a flex child refuses to shrink below its content's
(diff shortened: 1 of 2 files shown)


langyo:
Environment note: lint-commits check stuck in runner queue. Local verification green: vue-tsc --noEmit 0 errors (both commits), cross-verified by review round (fence-skip follow-up included in b6f65d6).

PR #245, opened 2026-08-21
✨ Add the shared one-shot health-endpoint probe primitive.

langyo:
## Summary
P59-W2 (hikari upstream): the last shared primitive this wave needs.

- **utils/healthProbe**: `probeHealthEndpoint(url, { timeoutMs, fetchFn })` — one fetch, one `AbortSignal.timeout`, one canonical minimal body type (`HealthProbeBody` with index signature). Never throws; returns `{ ok, latencyMs, body, reason }` with a `TimeoutError`-aware reason. Pure function — callers bring the URL, no app config.
- Replaces the per-app hand-rolled `/api`/health` fetches whose timeout/body policies drifted apart (chest `useHealthProbe`, plana-legacy `useEngineHealth`, erp's status poll — the latter two land in the companion PRs).

## Test plan
- `vue-tsc --noEmit` clean; `vitest run` **250/250**.
- Consumed downstream in shittim-chest #407 and erp #57 (merge this first).

PR #246, opened 2026-08-21
✨ Mirror actionable toasts to OS notifications while the page is hidden.

langyo:
## Summary
P59-W5 — closes the scan's "toast vs messaging" unification item at the right layer: **one visibility policy for both notification entry points**.

Direct `useToast` callers (chest's error handlers, the RPC layer) were silently dropped when the user was in another tab — only `useMessaging` callers got browser-notification routing. Now `push()` mirrors actionable severities (error/warning/success) through the browser transport while the page is hidden, with the same permission gate the router uses:

- visible page → toast only (unchanged); hidden + permission granted + actionable → toast + OS notification (identical to the messaging path).
- messaging-routed payloads still fire exactly one notification: the toast transport pushes through a suppression guard, so the router's own browser fire is never duplicated.
- `browserTransport`'s cron import goes straight to the runtime bus instead of the self-barrel, breaking the new useToast→browserTransport cycle.

## Test plan
- `vue-tsc` 0; `vitest` **256/256** (6 new: hidden-mirror, visible-no-mirror, info/loading-never-mirror, router-no-double-fire + toast-still-lands, permission-denied no-op, visible helper no-op).

### DIFF
--- a/packages/vue/src/composables/messaging/browserTransport.ts
+++ b/packages/vue/src/composables/messaging/browserTransport.ts
@@ -1,5 +1,5 @@
 import { defaultDurationFor, defaultRequireInteraction, type MessagePayload, type MessageTransport } from "./types";
-import { scheduleCronAfter } from "@celestia-island/hikari";
+import { scheduleCronAfter } from "../../runtime/cronBus";
 
 /**
  * Browser Notifications API transport.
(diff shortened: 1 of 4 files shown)


langyo:
CI 豁免记录（P59-W5）：Clippy (nightly) 在 Setup toolchain 步失败（nightly 工具链下载，环境性，#245/#244 同款）；hosted fmt+clippy+test ✅ + validate ✅（Rust 侧全绿——本 PR 仅触碰 TS）。本地门禁：vue-tsc 0 + vitest 256/256（6 新测试钉死镜像语义：hidden 镜像/visible 不镜像/info-loading 不镜像/路由不双发+toast 仍落地/无权限静默/可见 helper no-op）。

PR #248, opened 2026-08-21
✨ Add fixed-size month and year views with registered drill animations to the date pickers.

langyo:
## What

- `HkDatePicker` gains the days → months → years drill-down that `HkDateTimePicker` already had (clickable header title, back button, year chevrons), with the footer and time row kept visible in every view.
- Both pickers now render inside a **fixed-size stage** pinned to the day-view geometry (`HkPickerPane.scss`): opening a month/year grid no longer resizes the popup or inline panel — width and height stay identical.
- View switches animate through **direction-aware keyframes** (`hk-picker-pane-{fwd,back}-{in,out}`) registered in the CSS animation registrar (`registerAnimations.ts`); reduced-motion and the global `html[data-css-animations]` switch collapse them to an instant swap so a pane can never be stuck invisible.
- `HkDateTimePicker` time-stepper bumps no longer yank the user out of the drilled month/year grids (only real day changes re-anchor the view); focus moves to the incoming pane's primary control after a drill.
- The pickers now read their strings from the `hikari::` namespaced JSON bundles (matching the newer components), and all new keys (back / prev-year / next-year / prev-years / next-years) are added across all eleven locales in both the TOML sources and JSON bundles.

## Verification

- `vue-tsc --noEmit` clean; full vitest suite 257/257 green (incl. new drill-view, stepper-keep-view, stage-persistence tests).
- Browser-measured (Chromium): panel 280×293 / stage 250px identical across days/months/years views for both pickers; enter/leave animations observed running (`hk-picker-pane-fwd-in`/`-out`), global switch produces an instant swap.
- Two subagent review rounds completed (round-1 findings — header height missing from stage calc, TS strict errors, stepper view-snap, 16px slide overflow, tautological test — all fixed).


langyo:
CI note: GitHub-hosted fallback (fmt + clippy + test) passed 56s on the current head f1f06db. Self-hosted Run Tests / Coverage / Commit Message Lint are queue-stalled (>9 min on the shared runner pool) — environment-exempt per workspace CI policy; local verification is green: vue-tsc clean, vitest 257/257 (fork pool via --pool=threads), sass compiles, commit message validated by the local commit-msg hook. Merging.

langyo:
lint-commits environment note: the job failed with `fatal: Invalid revision range b1239b0..f1f06db` — the runner's bare-mirror checkout does not have the pre-force-push base SHA (documented mirror-sync gap). The commit message itself passes the identical celestia-devtools linter locally (the commit-msg hook enforced it at commit time). Hosted CI (fmt+clippy+test) green, local 257/257 + vue-tsc + sass green, two subagent review rounds approved. Merging under the environment-exemption policy.

PR #249, opened 2026-08-21
✨ Add the adaptive modal-drawer dialog shell and blocking toast confirmations.

langyo:
## Summary

- **HkAdaptiveDialog** (exported as `HAdaptiveDialog`): one `v-model` dialog contract that renders HkModal centered on desktop and HkDrawer `side="bottom"` on mobile (switched live by `useBreakpoint().isMobile`). Props `title/closable/width/mobileSize/footerActions/panelClass`, slots `default/header/footer` normalized across both shells (footer slot wins over footerActions; mobile renders the same actions as an HButton row). Overlay, teleport, z-index stacking, body scroll lock, focus trap and Escape stay delegated to the inner components via usePopupManager — nothing hand-rolled.
- **HkBlockingToast + useBlockingToast**: a promise-gated confirmation primitive — toast-like presentation (compact card in the toast column, kind "toast" in the popup registry) but flow-blocking (no auto-dismiss; `showBlockingToast(message, opts)` resolves true/false only on explicit choice; optional `timeoutMs` resolves false). Built for consent gates such as group-join visibility disclosure.
- **HkDrawer** gains an additive `afterLeave` emit (mirrors HkModal) so the adaptive shell forwards a uniform leave signal.
- i18n keys `hikari::blockingToast.{confirm,cancel}` added in all 11 locales (reusing each locale's confirmDialog wording).
- Bump `@celestia-island/hikari` 0.4.25 → 0.4.26.

## Verification (3-round cycle)

- R1 adversarial verify: PASS-WITH-NITS → fixed `mobileSize` default (85vh was dead — HkDrawer clamps bottom sheets to 70vh) and documented the mid-flip `afterLeave` caveat.
- R2 fresh-eyes verify: PASS-WITH-NITS → applied the in-PR version bump.
- R3 final: `vue-tsc --noEmit` clean, full vitest suite 29 files / 268 tests green (18 new), lint green.

First consumer: the shittim-chest voice-engine install wizard (centered modal on desktop, bottom drawer on mobile).

## Notes

- The blocking toast intentionally paints above modals (existing `--hk-z-toast` convention) so consent gates stay answerable over any dialog stack.
- `pnpm` in worktrees with symlinked node_modules needs `--config.verify-deps-before-run=false` (pnpm 11 would otherwise attempt an implicit install).

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.4.25",
+  "version": "0.4.26",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library \u2014 production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkAdaptiveDialog.scss
+++ b/packages/vue/src/components/HkAdaptiveDialog.scss
@@ -1,0 +1,22 @@
+// HkAdaptiveDialog — only the seams the two shells cannot provide
+// themselves: the composed drawer header (title + sub-header stacked in
+// HkDrawer's header slot, which otherwise replaces the title) and the
+// footerActions HButton row rendered inside HkDrawer's footer (HkModal
+// already renders that row itself; HkDrawer has no footerActions prop).
+// Everything visual (overlay, panel, transitions, footer padding and
+// divider) comes from HkModal.scss / HkDrawer.scss.
+
+.hk-adaptive-dialog-drawer-header {
+  display: flex;
+  flex-direction: column;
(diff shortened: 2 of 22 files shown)


langyo:
Merging over environmental CI state per workspace CI policy (§7): the only required check `lint-commits` has been queued >25 min with no self-hosted runner assigned (pool saturated by other repos' runs); the hosted `fmt + clippy + test` job PASSED (1m1s). Local evidence: commit subject passes `celestia-devtools commit-msg-lint`; vue-tsc build, full vitest suite (29 files / 268 tests), and lint all green in the worktree.

PR #250, opened 2026-08-22
🌐 Migrate the dead hk.* i18n keys and localize the remaining hardcoded component defaults.

langyo:
## What

Completes the hikari i18n sweep: every component now resolves strings through the live `hikari::…` namespace instead of the dead `hk.…` prefix, and the remaining hardcoded English defaults are localized.

## Why

A previous namespace migration left ~40 call sites using the dead `hk.` prefix — those always fell back to the hardcoded English fallback, so non-English locales never saw translations. Additionally several `hikari::` keys referenced by components were missing from the locale bundles, and a few components hardcoded English defaults outright.

## Changes

- **Key migration**: all `t("hk....")` call sites → `t("hikari::…")` across HkSearchInput, HkAvatar, HkModal, HkDrawer, HkConfirmDialog, HkKeywordSearchModal, HkZoomToolbar, HkLogo, HkMediaControlBar, HkScrollContainer, HkAlert, HkTable, HkMinimap, HkErrorBoundary, HkBreadcrumb, HkToast — same English fallbacks, zero behavior change for en.
- **New bundle keys**: `zoomToolbar.*`, `mediaPlayer.*` (incl. `fullscreen`), `auth.solvingChallenge`, `secret.*`, `theme.*`, `tablePage.emptyTitle`, `adminShell.navTitle`, `menu.back`, `toast.msgs`, `statusPill.ms`, `protocol.decline`, and the new `adminHeader.*` labels — added to all 11 locales in both `components.json` (`hikari::`) and `components.toml` (`hk.`).
- **Hardcoded-default fixes**: HkAdminHeader (7 labels), HkAdminShell navTitle, HkMenu back aria-label, HkToast count suffix, `useConfirm` defaults, HkSplash alt, HkStatusPill " ms" unit — all via `t("hikari::…", fallback)` with the fallback equal to the previous literal.
- Drive-by: fixed a pre-existing `FakeNotification implements Notification` typing error in `toastMirror.test.ts` (from #246) that failed `vue-tsc --noEmit` on master.

## Verification

- `vue-tsc --noEmit` — clean
- `vitest run --pool=threads` — 281/281 tests pass (30 files)
- `sass --no-source-map src/styles/index.scss` — fails identically on clean master (missing `variables` load path for the standalone invocation); no scss touched in this PR
- No `t("hk.` call sites remain in `src`

## Note on translations

[...]

PR #251, opened 2026-08-22
🔧 Route component timers and overlays through the shared animation, interval and overlay contexts.

langyo:
Routes component timers and overlays through hikari's shared animation/interval/overlay/popup contexts so reduced-motion, hidden-tab and z-band behavior is uniform.

- **HkPasswordInput**: ripple canvas loop via `onFrame(..., "normal")` (parks under reduced motion), autofill poll via `scheduleInterval` (hidden-tab aware), one-shot rAFs via `onceFrame`, flash/focus-reclaim timeouts via `scheduleCronAfter` (must fire even when the rAF bus is parked).
- **useConnectionProbe**: 15s retry window (`scheduleIntervalAfter`) and 1s countdown ticker (`scheduleInterval`) — the intervalBus designed case (data polling, not animation).
- **HkNumberInput**: press-and-hold 300ms delay + 50ms repeat via `scheduleAfter`/`scheduleEvery` (animationBus), `stopHold` semantics kept.
- **HkCheckbox**: `animTimer` now cleared on unmount and runs on a cronBus one-shot (always clears, even under reduced motion).
- **HkRollingNumber**: 350ms cronBus fallback commit applies the final digit when `animationend` never fires (keyframes paused at 0% under `data-css-animations="0"` / reduced motion); cancelled on `animationend` or unmount.
- **useOverlay**: registry keyed by per-instance unique ids instead of display name — same-named instances no longer overwrite each other; name/group metadata retained (`isOverlayOpen`/`closeAll`/group semantics unchanged, API backward compatible).
- **HkSelect / HkColorPicker / HkPopupSelect**: popouts register with `useOverlay` on open and unregister on close/unmount so `closeAll()`/`isOverlayOpen()` see them (HkPopupSelect keeps its legacy module singleton).
- **useToast**: per-slot cap (default 5, oldest dropped) and consecutive-duplicate dedupe (same slot + same text replaces instead of appends, dismiss clock restarts).
- **HkToast**: stack registers with `usePopupManager` (kind `"toast"`) — participates in the unified z-band ladder.
- **HkTooltip**: registers with `usePopupManager` (kind `"tooltip"`); SCSS z fallback fixed to `var(--hi-z-tooltip, 10000)` (theme.scss defines 10000 at `:root`).
- **HkMenu**: Escape closes the whole cascade (desktop panels + mobile sheets), mirroring HkPopover.
- **HkDrawer**: focus trap + focus restore mirroring HkModal (`focusFirst`/`trapFocus`).

Tests added: useOverlay unique-id registry, useToast cap/dedupe, HkRollingNumber fallback commit. Full vitest suite green (33 files / 295 tests). `vue-tsc --noEmit` shows zero new errors — one pre-existing TS2416 in `toastMirror.test.ts` (verified identical at baseline HEAD with changes stashed; file outside this PR's scope).

### DIFF
--- a/packages/vue/src/components/HkCheckbox.tsx
+++ b/packages/vue/src/components/HkCheckbox.tsx
@@ -1,6 +1,7 @@
 import { Check } from "lucide-vue-next";
-import { defineComponent, ref, type PropType } from "vue";
+import { defineComponent, onUnmounted, ref, type PropType } from "vue";
 
+import { scheduleCronAfter, type CronHandle } from "../runtime/cronBus";
 import HkLabel from "./HkLabel";
 import "./HkCheckbox.scss";
 
@@ -24,12 +25,16 @@ export default defineComponent({
   },
   setup(props, { emit, slots }) {
     const animating = ref(false);
-    let animTimer: ReturnType<typeof setTimeout> | null = null;
+    let animTimer: CronHandle | null = null;
 
     function markActive() {
       animating.value = true;
-      if (animTimer) clearTimeout(animTimer);
-      animTimer = setTimeout(() => {
+      // cronBus one-shot (not the rAF-driven animationBus one): the
+      // flag must always clear, even while the animation bus is parked
+      // under reduced motion.
+      animTimer?.disconnect();
+      animTimer = scheduleCronAfter(() => {
+        animTimer = null;
         animating.value = false;
       }, 300);
     }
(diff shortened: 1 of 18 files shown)


langyo:
CI note: some self-hosted jobs failed at the checkout step (bare-mirror sync for the fresh branch — documented infra gap); lint-commits passed. Local verification green after rebasing onto master-with-#250: vue-tsc clean + vitest 295/295 (33 files). Hosted fallback result pending at merge time. Merging per environment-exemption policy.

PR #254, opened 2026-08-22
🐛 Fix the animation bus per-entry timing and make the unified overlay close and z ladder truthful.

langyo:
## Design rationale

Three follow-up fixes balancing performance (throttled priorities) with fine-grained animation control (refresh-rate-invariant, clamped timing).

### 1. animationBus per-entry delta
Normal (33ms) and idle (2000ms) entries previously received the shared per-tick `delta` (~16.7ms at 60Hz), so phase-advancing callbacks ran at ~0.5x speed at 60Hz and ~0.2x at 120-144Hz. HkPasswordInput's ripple sweep and HkPlaceholderMarquee were affected (the marquee's speed-invariance comment was false).

Now each normal/idle callback receives `min((now - lastRun) / 1000, MAX_DELTA)` computed before `lastRun` advances — the exact real elapsed time since the entry's OWN last run, capped at 100ms so a jank/hiccup spike stalls the animation instead of teleporting it. The designed 0.3s ripple sweep now takes 0.3s at 60/120/144Hz; the marquee holds its px/sec speed. Sync entries keep the raw per-tick delta (every-frame precision tier). Priority docs now state the tradeoff: sync = per-frame fine control (most CPU), normal ≈30fps balanced, idle ≈0.5fps cheap — all delta-accurate.

### 2. useOverlay closeAll actually closes popouts
`closeAll()`/group-close only flipped the internal `isOpen` ref, so the visible popout (bound to the component's own open ref) stayed up. Entries now carry an optional `onCloseRequested` hook invoked alongside the internal flip, each wrapped in try/catch so one throwing callback cannot abort the rest (swallow + console.warn). HkSelect, HkColorPicker and HkPopupSelect wire it to flip their own open ref, which tears down the popout (their existing close paths unregister the popup manager). HkPopupSelect's legacy module singleton is untouched.

Also fixed the stale-registration NIT required for `isOverlayOpen` correctness: HkModal and HkDrawer registered once and only called `overlay.open()`, so entries stayed registered while closed-but-mounted. They now register/unregister with the open state (watch modelValue → open/close, unmount cleanup).

### 3. popup z-ladder reclaim
`nextZ` only reset when the registry emptied; HkToast's persistent registration (mounted forever) prevented the reset → the ladder drifted +2 per popup forever. Now on assignment the next z is derived from the max z of the CURRENTLY registered entries (reclaiming bands when popups unregister) and resets to the base when the registry empties — no register/unregister churn, HkToast stays as-is.

## Verification
- `vue-tsc --noEmit`: clean
- `vitest run --pool=threads`: 29 files / 267 tests green (one transient HkDateTimePicker transition-timing flake on first full run; passes consistently on rerun and on pristine master; unrelated to these changes)
- New tests: per-entry delta accuracy (16.7ms ticks → dt ≈ 0.033), clamp (500ms gap → dt ≤ 0.1), sync raw tick delta, idle clamp; closeAll onCloseRequested wiring + throw isolation + unique-id close; z-band reclaim (A→B→unregister B→C: C reclaims B's band, not B+2), empty-registry reset, scroll-lock balance.

## Files changed
- packages/vue/src/runtime/animationBus.ts
- packages/vue/src/runtime/useOverlay.ts
- packages/vue/src/runtime/usePopupManager.ts
- packages/vue/src/runtime/animationBus.test.ts (new)
- packages/vue/src/runtime/useOverlay.test.ts
- packages/vue/src/runtime/usePopupManager.test.ts (new)
- packages/vue/src/components/HkSelect.tsx
- packages/vue/src/components/HkColorPicker.tsx
- packages/vue/src/components/HkPopupSelect.tsx
- packages/vue/src/components/HkModal.tsx
- packages/vue/src/components/HkDrawer.tsx

### DIFF
--- a/packages/vue/src/components/HkColorPicker.tsx
+++ b/packages/vue/src/components/HkColorPicker.tsx
@@ -126,8 +126,12 @@ export default defineComponent({
 
     // Registered with the overlay registry so closeAll()/isOverlayOpen()
     // see the open popout; the popover inside handles z-stacking via the
-    // popup manager.
-    const overlay = useOverlay({ name: "hk-color-picker" });
+    // popup manager. The onCloseRequested hook makes a global closeAll()
+    // flip this component's own open ref, which tears the popover down.
+    const overlay = useOverlay({
+      name: "hk-color-picker",
+      onCloseRequested: () => { isOpen.value = false; },
+    });
 
     watch(isOpen, (open) => {
       if (open) overlay.open();
--- a/packages/vue/src/components/HkDrawer.tsx
+++ b/packages/vue/src/components/HkDrawer.tsx
@@ -104,6 +104,12 @@ export default defineComponent({
           handle.value = manager.register("drawer", true);
           overlayHook.open();
           previouslyFocused = document.activeElement as HTMLElement | null;
+        } else {
+          // Register/unregister WITH the open state so a closed-but-mounted
+          // drawer does not linger in the overlay registry (isOverlayOpen
+          // must reflect reality). The popup manager handle is torn down by
+          // the leave transition or by cleanup() on unmount.
+          overlayHook.close();
         }
       },
       { immediate: true },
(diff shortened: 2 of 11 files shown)


langyo:
CI note: heavy self-hosted jobs cancelled for runner economy (user directive); hosted CI passed 1m32s, lint-commits + PR title passed. Cross-verifier round APPROVE (4 NITs non-blocking). Local: vue-tsc clean + 267/267 tests. Merging.

PR #262, opened 2026-08-23
✨ Expose the embeddable color scheme editor and an external customize hook on the theme toggle.

langyo:
## Summary
Upstream foundation for the chest theme-customization dialog (colors / wallpaper / style tabs).

- **HkColorSchemeEditor** — the color-scheme editing form extracted from `HkColorSchemeDialog` as a standalone, embeddable component (props `initialDark`/`initialLight`/`initialGroups`, exposes `reset()` + `getDraft()`), keeping the full extension token-group surface (SCADA palettes, sections, hue-clamped pickers).
- **HkColorSchemeDialog** — now a thin modal wrapper with the exact same public API and behavior.
- **HkThemeToggle** — new `externalCustomize` prop + `open-customize` emit so host apps can own the customization entry point; entry relabeled via new `hikari::theme.customize` (all 11 locales).
- Version 0.5.0 → 0.5.1.

## Verification
- vitest: 35 files / 317 tests passed (existing dialog test unchanged).
- `vue-tsc --noEmit`: clean.
- lint: ok.

PR #264, opened 2026-08-23
🔥 Drop the dead auth protocol label rules and their customization classes.

langyo:
## Summary
`admin-tokens.scss` still shipped the `.s-auth-protocol-link/agree/row/text` rules from the old plana auth card. After the P57 de-customization the chest auth forms are their only consumer, and chest stopped using them — yet the leftover `.s-auth-protocol-text` rule kept overriding the HkLabel shell font size (12px vs the 14px default), which is exactly the mismatch this cleanup removes at the root.

## Changes
- Remove `.s-auth-protocol-link`, `.s-auth-protocol-link:hover`, `.s-auth-protocol-agree`, `.s-auth-protocol-row`, `.s-auth-protocol-text` from `packages/vue/src/styles/admin-tokens.scss` (35 lines).
- Version bump 0.5.0 → 0.5.1.

## Verification
- Grepped all family repos (shittim-chest / plana / arona / entelecheia / kirino): zero remaining references to the removed classes.
- `vue-tsc --noEmit` green (local).
- Companion chest PR drops the last usages of these classes from the auth views.

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.5.0",
+  "version": "0.5.1",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library \u2014 production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/styles/admin-tokens.scss
+++ b/packages/vue/src/styles/admin-tokens.scss
@@ -603,41 +603,6 @@
   color: var(--hi-color-muted, #666);
 }
 
-.s-auth-protocol-link {
-  color: var(--hi-color-primary, #58a6ff);
-  cursor: pointer;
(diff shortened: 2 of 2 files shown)


langyo:
Rebased onto master after #262 merged (8732134). The package.json bump now comes from #262's 0.5.1, so this PR's diff is only the 35-line dead-rule removal in admin-tokens.scss. Local verification: vue-tsc green; no family repo references the removed classes.

langyo:
CI status note: validate and the hosted fmt+clippy+test job pass; the self-hosted jobs (Format/Lint/tests/Coverage/lint-commits) sit in the shared runner queue (>15 min, known congestion). Local verification is green: vue-tsc 0 errors on both the hikari package and the chest webui that consumes it, and no family repo references the removed classes. Merging per the recorded environment-exemption policy.

PR #268, opened 2026-08-23
✨ Restore the connection status chain upstreamed from chest with compact mode and probe injection.

langyo:
## What

Wave 4 of the plana-legacy dissolution: the connection-status component chain — whose only live, enhanced copies lived in shittim-chest's plana-legacy layer — is upstreamed back into hikari (hikari's own copies were removed in #226 as unused in-repo).

- **HkStatusBar** — faithful port of chest's PlanaStatusBar including the unique **compact mode** (traffic-light dot + translated status text, version rows moved into the hover popover, versions carried in aria-label, `data-compact`), two-column grid version block, `fair`→Wifi quality icon, `Plug` protocol icon.
- **HkCountdownDigit** — self-contained port; `plana-*` classes/keyframes renamed `hk-*` (and registered in the animation registry, restoring the pair #226 removed).
- **HkConnectionStatus** — port with a **probe-injection overload**: hosts `provide(HK_CONNECTION_PROBE, { result, retryNow })` to swap in their own probe (e.g. an RpcClient-driven live transport tier); without it the component falls back to hikari's fetch-poll `useConnectionProbe`.
- **useConnectionInfo / ConnectionStateInput / HkConnectionInfo** re-exported from the barrel (they already lived in HkConnectionInfo.ts but were unreachable).
- `hikari::statusBar.*` keys in all 11 locale catalogs, values ported verbatim from chest's connection.json (including `{retryCount}/{maxRetries}` placeholders).

## Verification

- vue-tsc clean; vitest **349/349** (new HkStatusBar smoke test covers default + compact).
- Mechanical rename-diff against the chest sources shows byte-identity apart from the intended changes.

Consumer side (shittim-chest) switches to these exports and deletes the 13 dead copies in the chest PR.


langyo:
Round-1 verifier flagged the region-name regression (hardcoded Chinese-only map from the pre-upstream era); fixed in the follow-up commit: ported chest's locale-aware `regionDisplayName()` (political-name i18n keys first — zh-Hans/zh-Hant carry 中国大陆/中国台湾/中国香港/中国澳门 — then Intl.DisplayNames, raw code last).

PR #271, opened 2026-08-24
✨ Add an auto-collapsing window mode to the horizontal timeline.

langyo:
## What

Wizard-style steppers built on `HkTimeline` (e.g. the chest add-provider wizard's 4-step row) look broken on phones: the full row cannot fit the narrow width, labels collide and the connectors are starved.

This adds a **compact window mode**: when the full row cannot fit its host, the timeline collapses into a three-cell grid — previous / current / next — where:

- the current step stays dead-center and fully opaque,
- at most one neighbour shows per side,
- a side **fades toward its outer edge** when further hidden steps continue past it (the about-to-disappear hint), and shows no fade when the neighbour is the true first/last step,
- boundary positions render an empty side cell (first step → no left content; last step → no right content),
- connectors regain the horizontal space the crowded full row denied them.

## API

New `collapse` prop: `"auto" (default) | "always" | "never"`.

- `auto` measures the host with a ResizeObserver (sum of step widths vs host width, 16px expand hysteresis) and re-measures from a clean full render whenever labels change (locale switches).
- `always` forces the window (narrow cards, deterministic tests).
- `never` keeps the classic full row.
- Windowing only engages for horizontal rows with **more than 3 steps** (a 3-step row is its own window); vertical timelines are untouched.
- Window computation is exported as a pure `computeTimelineWindow()` for unit tests.

Backward compatible: existing consumers render identically on wide hosts.

## Verification (local, node-1)

- `vitest run` in packages/vue: **44 files / 370 tests pass** (12 new HkTimeline tests: window DOM structure, fade flags, boundary emptiness, connectors, ordinals, clickable neighbours, never/vertical/3-step exclusions).
- `vue-tsc --noEmit` clean.
- DemoApp gains a 280px-wide auto-window showcase plus a first-step `collapse="always"` sample.
- 0.5.1 → 0.5.2.

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.5.1",
+  "version": "0.5.2",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library \u2014 production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/DemoApp.vue
+++ b/packages/vue/src/DemoApp.vue
@@ -195,6 +195,14 @@
     <section>
       <h2>HTimeline</h2>
       <HTimeline :steps="timelineSteps" current-key="v2" />
+      <h3>Narrow host → auto window (prev / current / next, edges fade)</h3>
+      <div style="max-width:280px;padding:8px;border:1px dashed var(--hi-color-border, #888);border-radius:8px">
+        <HTimeline :steps="wizardSteps" current-key="vendor" />
+      </div>
+      <h3>collapse="always" at the first step</h3>
+      <div style="max-width:280px;padding:8px;border:1px dashed var(--hi-color-border, #888);border-radius:8px">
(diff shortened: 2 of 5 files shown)


langyo:
Verification record (3-round adversarial cycle, all rounds used subagents): R1 found a real layout defect (window-mode `[data-last]` cascade defeated the 1fr stretch — after-side fade masked empty space, connector was a stub; plus auto-measure summed shrunk flex rects and never collapsed in a real browser) — fixed in 8d693f9 (window call sites never set data-last; neighbours own the connectors; natural width measured via offscreen max-content clone), verified live in a real browser at 320px/360px/420px: auto-collapse fires, after cell stepW==cellW with 31px connector + fade mask applied, first/last-step windows have the empty side cell, wide host stays full. R2+R3 APPROVE (44 files/370 tests, vue-tsc clean, reactivity/leak/hysteresis/grid-centering all audited). Known non-blocking: pre-existing flaky test in HkDateTimePicker.test.ts (~50ms timing race, fails ~2/5 runs under node-1 load, byte-identical to master, untouched by this PR) — environment/flake, not merged over.

PR #283, opened 2026-08-25
✨ Route the back gesture through open windows before page history.

langyo:
## Summary

Window-first back priority for every window-like surface: the mobile
back gesture and the desktop browser back button now close the topmost
open window (modal / drawer / menu sheet) first; only when every window
has closed does the gesture reach page history and actually navigate.

## Implementation

- **`runtime/backStack.ts`** — central back-guard service generalizing
  HkMenu's Android modal-navigation mode: every window pushes marked
  history entries while open; ONE module-level popstate listener
  dispatches each gesture to the topmost open window (**open order, not
  mount order**); programmatic closes rewind via suppressed traversals.
- **`HkModal` / `HkDrawer`** — integrate via a `backGuard` prop
  (default on) gated by `closable`: non-closable surfaces never own
  history; mid-flight `closable` flips sync their entries.
- **`HkMenu`** — migrates onto the shared service, multi-level sheet
  semantics preserved; the sidebar variant now destroys its guard
  (fixes a record/listener leak).
- **Traversal discipline** — rewinds are deferred + re-validated at
  flush time (same-tick close-then-open never fires a stale traversal);
  `destroy` rewinds synchronously (unmount sweeps precede sibling
  mounts); a traversal issued with no listener skips suppression
  bookkeeping so a stale counter can never eat a later genuine gesture.

## Verification

- 392 vitest tests green (12 new service unit tests incl. open-order
  dispatch, same-tick close-then-open, unmount-mid-traversal, leak
  hooks; modal/drawer back-close coverage), vue-tsc clean, two
  consecutive full runs.
- Real-browser E2E (Chromium, 390px viewport): modal back-close keeps
  the page; nested modals peel one level per back; X-close leaves zero
  dead entries; router push while open then close leaves the router
  entry intact; drawer back-closes.

Closes the window-first back priority directive (PLAN P71).

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.6.2",
+  "version": "0.6.3",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkDrawer.test.tsx
+++ b/packages/vue/src/components/HkDrawer.test.tsx
@@ -1,0 +1,73 @@
+import { afterEach, describe, expect, it } from "vitest";
+import { createApp, defineComponent, h, nextTick, ref } from "vue";
+
+import HkDrawer from "./HkDrawer";
+
+const mounts: ReturnType<typeof createApp>[] = [];
+const containers: HTMLElement[] = [];
+
+afterEach(async () => {
+  for (const app of mounts.splice(0)) app.unmount();
+  for (const el of containers.splice(0)) el.remove();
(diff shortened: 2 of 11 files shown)


langyo:
CI note: remaining pending jobs are self-hosted Rust queue backlog (runner saturation). This PR touches only `packages/vue` TypeScript sources + package.json version — no Rust code. Hosted `fmt + clippy + test` ✅, `lint-commits` ✅, `validate` ✅. Local verification: 392 vitest green (×2 consecutive runs), vue-tsc clean, real-browser E2E 5/5 (modal back-close, nested peel, X-close no-dead-entry, router-foreign-top coexistence, drawer). Merging per PLAN §7 CI policy (environment-only queue failures exempt with local verification green).

PR #284, opened 2026-08-25
🔧 Show only the status dot in compact connection footers.

langyo:
HkStatusBar compact mode drops the translated status text — mobile footers show the bare traffic-light dot; tapping it still opens the popover with full connection details (latency/tier/versions), and the aria-label keeps the state for assistive tech. Test updated.


langyo:
Merged with environment-only exemptions recorded: Test (nightly/stable) and Test Coverage still pending on the saturated self-hosted queue (>15 min); lint-commits pending in queue. Local verification green: vitest HkStatusBar 2/2, fmt + clippy + test (hosted) PASS.

PR #285, opened 2026-08-25
🐛 Keep the back-guard rewind claim through same-tick normalize pushes.

langyo:
## Summary

Follow-up to #283: fix the N1 normalization drift and document the
remaining dead-marker trade-off.

## What

- **`push()` advances `desired` by one level instead of snapping it to
  `count`.** The HkMenu normalizer's desktop→mobile flip runs
  `release()` (desired=0, rewind deferred) + `push()` in the SAME tick;
  snapping desired back to count made the deferred flush see
  `count == desired` and skip the rewind, stranding the stale entry —
  one back turned into two. A relative `desired++` keeps the rewind
  claim alive so the flush correctly rewinds to exactly one entry.
- **flush candidates use a `Set`** so a record present in both the
  rewind queue and the registry (normalize release-then-push) is
  visited once (behaviour-identical, removes the duplicate visit).
- **Head comment documents the dead-marker trade-off** (N2): a marker
  buried mid-stack when a router pushed above the window cannot be
  deleted by the history API; it is inert and self-releasing, at the
  cost of one extra back until the router re-syncs from the URL.

## Verification

- 393 vitest green (new N1 regression test asserts exactly one entry
  after a same-tick release-then-push and a single back closing the
  surface), vue-tsc clean.
- HkDateTimePicker flake under full-suite NFS load is unrelated and
  passes 22/22 in isolation (also reproducible on clean master).
- N2/N3/N4 intentionally left unchanged: N2 is a history-API
  limitation (removing it needs hooking global pushState for direction
  detection, or breaks forward history), N3 is baseline-shared, N4 is
  B1's defensive code.

Closes P71 续 (PLAN P71-d).

### DIFF
--- a/packages/vue/src/runtime/backStack.test.ts
+++ b/packages/vue/src/runtime/backStack.test.ts
@@ -174,6 +174,30 @@ describe("createBackGuard", () => {
     b.destroy();
   });
 
+  it("normalizes a same-tick release-then-push to exactly one entry (N1)", async () => {
+    // HkMenu's normalizer: a desktop→mobile flip releases the desktop
+    // residue (deferred rewind) and pushes a fresh root in the SAME
+    // tick. The rewind claim must survive the push, or the stale entry
+    // strands and one back turns into two.
+    const onBack = vi.fn();
+    const g = createBackGuard({ onBack });
+
+    g.push(); // mobile root
+    g.release(); // desktop flip: rewind deferred
+    g.push(); // back to mobile, same tick
+
+    await settle();
+    expect(g.entries).toBe(1);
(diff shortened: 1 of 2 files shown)


langyo:
CI note: change touches only `packages/vue` TypeScript (`backStack.ts` + test) — no Rust. Hosted `fmt + clippy + test` ✅ (1m44s), `validate` ✅, `lint-commits` verified locally exit 0. Remaining pending jobs are self-hosted Rust queue backlog (runner saturation, environment-only per PLAN §7). Merging with local verification green: 393 vitest (×2, DateTimePicker flake isolated 22/22), vue-tsc clean, plus the review subagent's mutation experiment confirming the N1 regression test distinguishes the fix.

PR #293, opened 2026-08-26
✨ Center panel loading spinners on both axes by default.

langyo:
## Summary
`HkSpinner` `center` mode becomes the shared loading-state placement contract for every panel/page/modal surface:

- Column flex, centered on **both axes** (`justify-content:center` + `align-items:center`)
- Fills flexible parents (`flex: 1 1 auto`, `align-self:stretch`, `width:100%`) and claims definite-height block parents via `height:100%`; a **fixed `8rem` min-height floor** (predictable on every viewport height) keeps auto-height parents placing the circle mid-panel instead of top-centered — previously `center` only horizontally centered with vertical padding
- Optional `minHeight` prop emits `--hk-loading-min-height` for sites wanting a different footprint; blank strings are ignored
- New component tests incl. default-floor negative paths; version bump 0.7.0 (minor: new prop + default behavior flip)

Downstream `<HSpinner center />` call sites (~40 across shittim-chest and erp.celestia.world) adopt this automatically; companion PRs there only clean hand-rolled wrappers.

Note: `center + text` now stacks text below the wheel instead of beside it (standard panel-loading arrangement).

## Verification
- `vitest run`: 416 passed (incl. HkSpinner suite)
- `vue-tsc --noEmit`: clean

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.6.6",
+  "version": "0.7.0",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.6.5",
+  "version": "0.6.6",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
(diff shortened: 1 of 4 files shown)


langyo:
Merge-gate exemption per PLAN §7 / AGENTS.md §7 CI policy: `validate` SUCCESS; `lint-commits` queued >60min without starting (self-hosted runner backlog). Exemption grounds: every commit on this branch passed the local `commit-msg` hook (`core.hooksPath=/mnt/codespace/_hooks`) which runs the same celestia-devtools commit-msg-lint; subjects independently re-checked by the R3 verifier. Local gates: vitest 416/416, vue-tsc clean.

PR #295, opened 2026-08-26
✨ Add a floating jump control and rework autoscroll follow.

langyo:
## Summary
- New `HkFab`: Material-style floating action button with optional speed dial, container-relative (`absolute`) or global (`fixed`) positioning, safe-area-aware insets + offset props, size/variant props, registry-based icon resolution (no lucide tree-shaking regression), capture-phase Escape / outside-pointer dismissal.
- `HkModal` autoFollow rewritten: settle-frame pinning via direct `scrollTop` assignment (shared animation bus) replaces smooth-scroll chains — streaming appends can no longer flash the follow state mid-flight; a sticky gesture flag (`wheel`/`touchstart`/`touchmove`/scroll-keys/`pointerdown`, restamped by pressed `pointermove`) ensures only real user motion cancels the tail; pending pins yield to live gestures under a 250 ms quiet window and drain once cancelled instead of fighting the finger.
- Semantics inverted per report: the "Auto" chip now renders only while actively following at the bottom; scrolling away swaps it for an `HkFab` jump-to-latest button that resumes follow on tap. The old inverted bar is removed.
- `HkScrollContainer` gains opt-in `followAffordance` exposing the same jump control for pinned log panes; existing consumers unaffected (default off).
- i18n: `hikari::modal.jumpToLatest` added to all 11 locales; version 0.6.5 → 0.7.0.

## Motivation
Mobile bug report on shittim-chest chat-log windows: the auto marker flickered during scroll, appeared when follow was cancelled (inverted), and no affordance existed to return to the tail. Desktop shares the code path.

## Verification
- vue-tsc 0 errors; vitest 424/424 across multiple consecutive full-suite runs (incl. 9 HkFab unit tests + 4 behavioral HkModal autoFollow regression tests; the slow-drag case demonstrably fails against the pre-fix algorithm).
- Adversarial review rounds traced fling/cancel ordering, teardown symmetry, hover-vs-press stamping, windowed-mode pinning; residual known limits documented (still-held drags may admit one pin before the next move re-stamps — self-healing).

Depends-on consumer wave: celestia-island/shittim-chest PR wiring `followAffordance` lands after this merges.

### DIFF
--- a/packages/vue/src/components/HkFab.scss
+++ b/packages/vue/src/components/HkFab.scss
@@ -1,0 +1,159 @@
+// HkFab.scss — floating action button + optional speed dial.
+// Attribute-driven positioning: data-positioning (absolute|fixed) +
+// data-corner select the anchored insets; --hk-fab-offset-* extends the
+// safe-area-aware base per instance (set from offsetX / offsetY).
+.hk-fab {
+  position: absolute;
+  z-index: var(--hk-fab-z, 40);
+  display: flex;
+  align-items: center;
+  gap: 0.5rem;
+  pointer-events: none;
+
+  &[data-positioning="fixed"] { position: fixed; }
+
+  // Layout axis follows the expansion direction (DOM order [actions, main]).
+  &[data-expand="up"] { flex-direction: column; }
+  &[data-expand="down"] { flex-direction: column-reverse; }
+  &[data-expand="left"] { flex-direction: row; }
+  &[data-expand="right"] { flex-direction: row-reverse; }
+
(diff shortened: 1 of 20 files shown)


langyo:
Exemption note (§7 CI policy): Clippy (nightly) failed at the Setup toolchain step (bootstrap/exit 1, zero lint output) — the documented shared-rustup environment failure class on these self-hosted runners, not a code-level result. Reference gates green locally (vue-tsc 0 errors, vitest 424/424 across consecutive full-suite runs incl. new regression tests) and hosted stable fmt+clippy+test PASS on this PR.

PR #296, opened 2026-08-27
🐛 Spread fallthrough attrs onto the icon button root.

langyo:
## What

`HkIconButton` sets `inheritAttrs: false` but never spread `$attrs` onto its root `<button>`, so every `aria-label` / `title` passed by a consumer was silently dropped — the buttons rendered, worked, but had no accessible name and no tooltip. Live victims include chest's 3D detail-panel close button (`PhysicalPreview.tsx`) and the new panel-config edit affordance.

The fix is one line (spread `{...attrs}` on the root button) plus a three-case regression test pinning the contract (fallthrough works; the declared `click` emit still fires; disabled + size classes compose with attrs).

## Verification

- `vitest run`: 53 files / 435 tests green (431 baseline + 3 new − net 435 with the file counted once)
- `vue-tsc --noEmit`: clean

### DIFF
--- a/packages/vue/src/components/HkIconButton.test.tsx
+++ b/packages/vue/src/components/HkIconButton.test.tsx
@@ -1,0 +1,75 @@
+import { afterEach, describe, expect, it } from "vitest";
+import { createApp, defineComponent, h } from "vue";
+
+import HkIconButton from "./HkIconButton";
+
+/**
+ * HkIconButton contract tests:
+ * - extra attrs (aria-label / title / data-*) fall through onto the root
+ *   <button> (the component sets inheritAttrs:false and MUST spread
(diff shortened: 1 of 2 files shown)


langyo:
CI exemption record (§7): all self-hosted failures die in the **Setup toolchain / Setup Rust** step (rustup download on the runner) — `Format`, `Check Formatting`, `Clippy (nightly)`, `Test (nightly)`, `Test Coverage`; none reached the actual fmt/clippy/test steps. The hosted `fmt + clippy + test` composite job **passes** (green, includes cargo fmt --check), `lint-commits` passes. Local verification: `vitest run` 53 files / 435 tests green (431 baseline + 3 new regression cases for attr fallthrough), `vue-tsc --noEmit` clean.

PR #298, opened 2026-08-27
✨ Let hosts retune floating surface finishes and sheet gaps.

langyo:
Host-side theming hooks requested from the chest style batch:

- **HkModal mobile sheet**: `--hk-sheet-bottom-gap` (was hard `bottom: 0`) and `--hk-sheet-footer-gap` — a docked phone popup can now sit close to the bottom edge without touching it; safe-area inset stays stacked on top so home-bar devices never lose tappable footer chrome.
- **HkPopover glass layer**: `--hk-popover-bg` / `--hk-popover-blur` with the historical look as fallback defaults.
- Source-contract tests pin both hook sets (`HkModal.sheetgap.test.ts`, `HkPopover.surfacevars.test.ts`) so the values cannot silently regress to hard-coded ones again.

Chest consumption lands in the companion shittim-chest PR (solid / glass blur(2px) / acrylic blur(4px)+saturate surface finish wired through `html[data-surface]`).

### DIFF
--- a/packages/vue/src/components/HkModal.scss
+++ b/packages/vue/src/components/HkModal.scss
@@ -310,7 +310,9 @@
     max-width: 100% !important;
     width: 100%;
     top: var(--hk-sheet-top-inset, 4.25rem);
-    bottom: 0;
+    // --hk-sheet-bottom-gap keeps the docked sheet from touching the very
+    // bottom edge (host-tunable; chest narrows the default visual gap).
+    bottom: var(--hk-sheet-bottom-gap, 0.5rem);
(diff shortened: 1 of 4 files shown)


langyo:
Env-only failures recorded per PLAN §6.2: all 6 failing jobs die at checkout (`upload-pack: not our ref`) or `Setup toolchain` (rustup) before compiling anything; no code-level check has run. Local verification green: targeted vitest 25 passed (new source-contract tests sheetgap/surfacevars + autofollow + adjacent component regressions); lint-commits pass.

PR #294, opened 2026-08-26
✨ Rebuild the theme menu mode group on HkSegmented and unify list row highlights.

langyo:
## Summary
Polishes the shared theme menu (HkThemeToggle) used by chest (desktop + mobile), erp.celestia.world and e.celestia.world:

1. **Uniform row highlights** — every theme row now has identical geometry: an invisible trailing slot mirrors the custom-theme delete button's footprint, so the active highlight is the same width with symmetric padding on all rows. The check glyph moved inside each pill.
2. **Mode group on a generic component** — the color-mode area is rebuilt on the shared HkSegmented button group (Auto | Light | Dark).
3. **Auto-mode altitude cover** — while Auto is active, the Light/Dark halves are covered by a passive solar-altitude readout strip in the segmented typography (sun/moon glyph + live altitude like "+32.5°"). It is not directly switchable, but pressing it falls back to manual mode selecting whichever side auto currently resolves to (day → light, dusk/night → dark). Manual mode shows the three normal segments; covered segments are disabled under the overlay.

New i18n string `hikari::theme.autoAltitudeTip` in all 11 locales. Consumer apps need no changes — props/emits/slots are untouched.

## Verification
- 3-round verify cycle with independent subagents: R1 PASS (0 major), R2 APPROVE (0 major), R3 **SHIP** verdict.
- Targeted vitest 5/5 (3 new cases: overlay visibility/press resolution, segment disablement in auto, row uniformity incl. custom delete rows); full package suite 414/414 (+3 over green master baseline 411); vue-tsc --noEmit exit 0.
- Locale JSON validated programmatically across all 11 languages (exactly one added key each).

## Notes
- Per §7.2 no manual build was run; local verification is green per §7 gate, deployment handled by malkuth watchers downstream.
- Follow-up candidate (upstream engine, out of scope here): useTheme's currentPeriod is computed once at module load and never re-ticked against resolved geolocation.

PR #299, opened 2026-08-27
✨ Stack compact timeline steps vertically and add the shared step flow container.

langyo:
## What

Two coordinated changes to unify step/wizard UX across consumers (shittim-chest migrates onto this in a follow-up PR):

### 1. HkTimeline compact "window" mode redesign
The old compact mode laid labels BESIDE numbers and row-reversed the right cell so its label sat LEFT of its number — confusing on phones. New behavior:
- three fixed equal slots always present; each rendered step stacks the 24px indicator circle on top with its label BELOW as normal horizontal text (centered, ellipsised)
- previous/next cells dimmed (`--hk-timeline-neighbor-opacity`, default 0.55) to read as "previous / upcoming"; clickable completed neighbours brighten on hover
- an overlay of link segments joins neighbouring node centers, colored by adjacent status; where hidden steps continue past an edge a gradient-fading hint dissolves toward that container edge (single fade mechanism — masks deleted)
- empty edge cells stay truly empty; RTL supported via logical insets + mirrored fade paint; `aria-current="step"` added in both modes
- full-row desktop rendering untouched

### 2. New `HStepFlow` component
Generic step container: optional HkTimeline header bound via v-model + a direction-aware sliding body (`out-in`, forward/back translateX ±24px) fed purely by named slots keyed by step key. Additive API: `steps`, `modelValue`, `hideTimeline`, `timelineClickable`, `stickyHeader`, `collapse`; scoped slot arg `{ key, index, direction }`. Unknown keys render an empty body silently. Also:
- `.hk-modal-footer:empty { display:none }` so conditional footer phases never flash a bare bordered strip
- version 0.7.1 → 0.8.0

## Verification
- 3-round independent adversarial verification (R1/R2/R3), final verdict MERGE-APPROVED (findings round-tripped and fixed: discriminating direction-resync regression test, valid scaleX(-1) RTL mirror instead of invalid filter)
- vitest: 57 files / 454 tests green; vue-tsc --noEmit clean
- consumer impact: additive only; HkTimeline props/emits/public pure function unchanged

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.7.1",
+  "version": "0.8.0",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkModal.scss
+++ b/packages/vue/src/components/HkModal.scss
@@ -290,6 +290,12 @@
   justify-content: flex-end;
   gap: 0.5rem;
   flex-shrink: 0;
+
+  // Conditional footer slots (phases without actions, close transitions)
+  // render nothing — never leave a bare bordered strip behind.
+  &:empty {
+    display: none;
+  }
 }
 
 // ------
--- a/packages/vue/src/components/HkStepFlow.scss
+++ b/packages/vue/src/components/HkStepFlow.scss
@@ -10,6 +10,21 @@
   --hk-stepflow-duration: 0.2s;
 }
 
+// Sticky header mode: keeps the step indicator visible while the body
+// scrolls inside long modal hosts. Surface tint + blur approximate the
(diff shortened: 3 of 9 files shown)


langyo:
CI exemption record (§6.2/§7): 'Check Formatting' failed in its **Setup Rust** step (dtolnay/rust-toolchain@nightly could not install — job log: Setup Rust=failure, cargo fmt skipped); 'Test (nightly)' is the same hosted nightly-toolchain class. This diff contains **zero .rs files** (verified `git diff origin/master...HEAD --name-only | grep .rs` = empty), and the same workflow family also fails on current master (`test.yml`=failure, `fmt.yml`=failure on merged #294/#298). Local verification (substantive gates): vitest 57 files / **454 tests green**, vue-tsc --noEmit exit 0, 3 independent adversarial review rounds R1/R2/R3 all PASS → MERGE-APPROVED. Self-hosted queue >15 min at merge time per runner-pool status; treating remaining pending as environmental.

PR #301, opened 2026-08-27
✨ Add a form slider component with a persistent themed thumb and steps.

langyo:
## Summary
- New `HSlider` form/settings slider in packages/vue, exported alongside the existing primitives:
  - Persistent, always-visible thumb (the media seek bar keeps its hover-only thumb by design; settings surfaces need a control that reads as draggable at first glance — reported as a bare progress bar with no knob in the theme dialog).
  - Thumb + fill follow the active theme via `--color-primary`.
  - `step` snapping (every emitted value lands on `min + n * step`), optional `showTicks` per stop with active-tick tinting.
  - Full keyboard support (arrows, PageUp/Down, Home/End) and slider aria semantics incl. `formatValue` → `aria-valuetext`.
  - `sm`/`md` sizes, disabled state, same window-level pointer-drag model as HkMediaSlider.
- Version 0.8.0 → 0.9.0.

## Verification
- `vitest run`: 58 files, 466/466 green (includes 9 new HkSlider tests; one pins the always-visible themed thumb as a source contract so the hover-gated regression cannot return).
- `vue-tsc --noEmit`: clean.
- Downstream: shittim-chest will consume this for the theme dialog DPI control + wallpaper brightness slider in a follow-up PR right after this merges.

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.8.0",
+  "version": "0.9.0",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkSlider.scss
+++ b/packages/vue/src/components/HkSlider.scss
@@ -1,0 +1,99 @@
+.hk-slider {
+  position: relative;
+  // Comfortable pointer/touch hit area; the visual track is drawn inside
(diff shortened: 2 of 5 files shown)


langyo:
Merging per §6/§7: hosted fmt+clippy+test PASS (code-level), local vitest 466/466 + vue-tsc clean, cross-verify subagent verdict PASS. Remaining self-hosted jobs queued >15 min = environmental exemption per §7 CI policy; lint-commits already validated locally by the repo commit-msg hook.

PR #304, opened 2026-08-28
✨ Add a sliding thumb to segmented groups and measured morphing tab indicators.

langyo:
Fixes the two reported regressions in one wave:

- **Segmented control lost its sliding-thumb animation** (#294 rebuilt the theme-toggle mode group onto `HkSegmented`, which had no thumb). `HkSegmented` itself now renders an animated sliding thumb: a measured highlight positioned with real geometry (`useMeasuredHighlight` composable), transitioning `transform`/`width` at 0.2s. Per-segment checked styling stays as the fallback until a non-zero width is measured (jsdom/SSR never get there), so there is never a flash of no-indicator.
- **Morphing-tabs indicator drifted with 4+ tabs**: the old `--tab-count`/`--tab-index` percentage math assumed equal-width triggers, but `flex: 1 1 0` + `min-width: auto` + `white-space: nowrap` lets overflowing labels widen some triggers (e.g. 颜色/壁纸/样式/DPI 控制 in a narrow bottom sheet). The indicator now uses measured trigger offsets/widths from the same composable.

## Details

- New public composable `useMeasuredHighlight({ container, activeIndex, itemSelector, extraSources? }) → { x, width, ready }`: measures on mount (nextTick), on active-index/extra-source changes, via `ResizeObserver` on the container (guarded), and once after `document.fonts.ready`. `offsetLeft` accumulation walks the offsetParent chain when the container is not the direct offsetParent. jsdom/happy-dom/SSR-safe: layout reads are 0 there, `ready` stays false, nothing throws, no NaN (finite-coerced). Disconnects on scope dispose. Exported from the package root.
- `HkSegmented`: thumb is the first root child (`aria-hidden="true"`), root gets `data-ready="true"` only once measured; root is `position: relative`, segments promoted to `z-index: 1` so text stays above the thumb; `data-ready` suppresses the per-segment checked background/shadow (checked text color kept); a disabled group dims its thumb like its labels; negative `modelValue` match clears the highlight instead of pointing at segment 0 (mirrors all-false `aria-checked`); `prefers-reduced-motion` disables the transition; touch form (≤767px) unchanged and thumb-correct.
- `HkMorphingTabs`: indicator carries inline `transform`/`width`, fades in on the triggers' `data-ready`, `aria-hidden="true"`, reduced-motion opt-out; teleport/a11y/footer behavior untouched.
- `HkThemeToggle.scss`: the auto-mode altitude cover strip gets `z-index: 2` so the promoted segments (z-index 1) don't paint above it.
- Version bump 0.11.0 → 0.12.0.

## Verification

- Scoped: `vitest run HkSegmented.test.tsx HkMorphingTabs.test.tsx HkThemeToggle.test.ts` → 18/18 (HkThemeToggle's 5 stay green).
- Full suite: `vitest run` → **59 files, 475 passed / 0 failed** (baseline before: 466 tests).
- `vite build` → success. (`vue-tsc` TS2688 'node' types is the known environmental failure, not run as a gate.)
- 3-round verify cycle with independent review subagents; fixed disabled-thumb dimming, no-match phantom highlight, mount grow/slide artifact + zero-width shadow sliver, and missing indicator `aria-hidden` along the way.

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.11.0",
+  "version": "0.12.0",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkMorphingTabs.scss
+++ b/packages/vue/src/components/HkMorphingTabs.scss
@@ -23,19 +23,42 @@
   background: color-mix(in srgb, var(--hi-color-surface) 60%, transparent);
 }
 
+/*
+ * Indicator geometry is measured (useMeasuredHighlight) and applied as
+ * inline transform/width, so unequal trigger widths (overflowing labels
+ * under `flex: 1 1 0`) no longer drift the highlight off the active tab.
+ * `left: 0` + translateX(x) lands on the trigger's own offsetLeft, which is
+ * padding-edge relative — the same reference frame as the old `left: 2px`.
+ * Width 0 + opacity 0 (before the first measurement) keep the indicator
+ * invisible with no mount grow/slide; `data-ready` fades it in once real
+ * layout exists.
+ */
 .hk-morphing-tabs-indicator {
   position: absolute;
   top: 2px;
-  left: 2px;
+  left: 0;
   height: calc(100% - 4px);
-  width: calc((100% - 4px) / var(--tab-count, 1));
+  width: 0;
   border-radius: var(--hi-radius-sm, 4px);
   background: color-mix(in srgb, var(--hi-color-primary) 12%, transparent);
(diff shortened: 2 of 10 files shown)


langyo:
Checks note: `lint-commits`, `validate`, `Verify version consistency` and the hosted `fmt + clippy + test` composite all pass. The self-hosted `Tests`/`Clippy`/`Run Tests`/`Coverage`/`Format` runs failed for documented environmental reasons only: (1) `upload-pack: not our ref` — bare-mirror sync lag on the fresh PR merge ref (checkout step, retried); (2) `Format`/`Coverage` fail in the runner's own `Setup toolchain` step (rustup: failure removing component 'cargo-x86_64-unknown-linux-gnu' on the self-hosted machine) before reaching the code. The diff touches only Vue/TS/SCSS; local verification is green: vitest 59 files / 475 tests passed, `vite build` success.

PR #305, opened 2026-08-28
✨ Repower HkMenu with the HkSelectPanel surface and slot extensibility.

langyo:
## What

HkMenu's private popover/sheet engine (own positioning, own mobile fullscreen-sheet stack, own back-guard wiring) is replaced by the existing shared detachable dropdown surface: every menu level — root and nested — now renders through `HkSelectPanel`. Menus and dropdowns are the same implementation. `HkSelectPanel.tsx` and its tests are untouched.

## New surface API (frozen contract)

- `title?: string` → passed to HkSelectPanel `title` (mobile sheet header text, desktop popout `aria-label`). Existed as a prop already; now actually reaches the shared surface.
- Slots `header` / `footer` → custom content rendered inside the panel surface above / below the item list, in BOTH the desktop popout and the mobile sheet (e.g. an avatar identity block pinned at the top of a user menu). Rendered at the root level.
- `selectMode?: "auto" | boolean` (default `"auto"`) → leading check column reservation. `"auto"` reserves when any item at the current level has `checked !== undefined`; boolean forces. `checked === true` renders ✓, the rest get an empty placeholder only while reserved — plain action menus show no column at all.
- `matchAnchorWidth?: boolean` (default `false`, shrink-to-fit) → passthrough to HkSelectPanel. Cascade placements (right-/left-start and sub-levels) always shrink-to-fit.
- Default slot → composition mode: raw consumer rows replace the `items` model (mutually exclusive per instance; no cascade in slot mode).
- Sub-levels: each child level is its own nested `HkSelectPanel` anchored to the parent row via a live synthetic anchor (desktop: popout right, flips left when out of viewport; mobile: one stacked sheet layer per level, one back-guard history entry each, grabber + `title` with the parent item label as fallback). Mobile rows get a >=44px touch min-height via the shared stylesheet.
- Everything else is kept: `items: HkMenuItem[]` (unchanged shape), `open` (v-model:open), `select(key, item)`, `anchorRef`, `placement` union incl. `right-start`/`left-start` (via synthetic cascade anchors), `offset`, `variant="sidebar"` (unchanged inline nav), menubar hover switching, keyboard Escape closing.

## Intentional behavior alignment (select-family parity)

- Breakpoint crossing while open now CLOSES the menu (HkSelectPanel's documented contract) instead of re-docking mid-flight.
- Escape closes the topmost panel level (surface-attached), not the whole cascade from anywhere.
- Mobile sheets use the shared grabber + title header; the old in-app back button is gone (back gesture / scrim / grabber pop one level). `hikari::menu.back` locale keys removed.
- `mobileBreakpoint` is accepted but now a documented deprecated no-op (the shared surface decides at the library-wide 768px breakpoint).
- Desktop cascade levels now each own one back-guard history entry (popouts are window layers in the shared model).

## Verification

- `vitest run`: 59 files, 487/487 green — HkMenu.test.ts rewritten to the new contract (23 tests: desktop cascade on `.hk-select-popout`, sibling click/hover switching, outside-click attribution through cascade popouts, check column auto/true/false, header+footer slots in both modes, slot composition mode, mobile sheet layers with back-guard pops, sheet title/grabber contract, breakpoint-crossing close, right-/left-start synthetic anchor cascade, sidebar suite intact). HkSelectPanel tests untouched and green (12).
- One review-blocked bug found and fixed during the verify cycle: click-attribution record was cleared on a microtask, which per the HTML microtask-checkpoint rule would run between listeners of one dispatch in real browsers (happy-dom hides this) — now cleared on a zero-delay macrotask.

## Version

packages/vue 0.12.0 → 0.13.0 (master already carried 0.12.0).


langyo:
**Environmental failure record (§6.2/§7 merge exemption)**

All currently failing checks fail BEFORE reaching any lint/test step, on shared self-hosted runner infra — not on this diff (which touches only `packages/vue`):

1. First wave: `fatal: git upload-pack: not our ref` on Checkout Project (the known §9.4 bare-mirror lag for a newly pushed ref). Evidence: jobs 98797996806/98797996523/98797996943 etc., run 33155755839.
2. After rerun (checkout now fine): `Setup Rust` fails with `error: could not parse settings file: '/home/lab/.rustup/settings.toml': TOML parse error at line 1, column 1` — the runner host's rustup settings are corrupted (concurrent-agent infra), failing Check Formatting / Lint / Run tests / Clippy ×2 / Format / Tests / Coverage identically at toolchain setup. Evidence: jobs 98800165738/98800166042/98800166006/98800169986/98800174585.

Checks that did run and PASS: `fmt + clippy + test (hosted)` (validates the tree), `lint-commits`, `validate`, `Verify version consistency`.

Local verification (authoritative per §7): `vue-tsc --noEmit` 0 errors; `vitest run` 59 files / 487 tests green, including the untouched HkSelectPanel suite (12). No Rust code is touched by this PR, and no failing job reached a compile or test step. Merging per §6.2/§7 with this record.

PR #306, opened 2026-08-28
🐛 Pin the leaving toast box so closing toasts slide out instead of flashing.

langyo:
## Problem

Closing a toast sometimes "flashed": the toast's height snapped tall, its width snapped to a sliver, and it vanished (or got flung sideways) instead of playing the entrance animation in reverse.

## Root cause (measured live in a browser against the real component)

The toast stack column is `position: fixed` + shrink-to-fit (`right` anchor + `max-width` only). When the **last** (or widest) toast starts leaving:

1. `*-leave-active { position: absolute; width: 100% }` takes the toast out of flow → the column instantly collapses to **zero width** → `width: 100%` resolves to a **32px padding-only sliver** (measured: 384px → 32px on frame 1).
2. The sliver's static position jumps to the collapsed edge, and the TransitionGroup FLIP `move` transition animates it back — flinging the deformed sliver ~400px across the screen while it fades (measured x: 1040 → 1439 over ~270ms).
3. When the message body keeps a nonzero width (stacked messages, count pill, a narrower sibling), the squeezed text rewraps — the tall/narrow flash users reported.

## Fix

- `pinLeaveGeometry` (new in `utils/dom.ts`): a before-leave hook freezes the leaving item's in-flow box — `right` (anchored to the containing block's **right** padding edge, the one coordinate that survives the collapse), `top`, `width`, `height` + inline `box-sizing: border-box`. All metrics come from the transform-free `offset*` family, read **before any style write** (interleaved reads once froze a transient content-box rewrap during development).
- `clearLeaveGeometry` wired to `leave-cancelled` so an aborted leave never leaves stale pins.
- Wired into both `HkToast` and `HkBlockingToast`; the `width: 100%` footgun is deleted from both leave-active rules.

## Verification

- **Live browser trace** (real component, per-rAF geometry sampling): leaving toast holds **384×83 constant**, slides **exactly 1rem right**, opacity 1→0 over the 0.3s curve, removed at ~320ms — for both the last-toast and widest-of-two cases. Before/after traces attached to the task log.
- `vitest`: 15/15 across `dom.test.ts` (new), `HkToast.test.tsx` (new), `HkBlockingToast.test.tsx` (pin regression added); full package suite green.
- `vue-tsc --noEmit` clean.
- Independent subagent review: PASS on all 7 checklist items (incl. Vue 3.5.40 source check that `onBeforeLeave` runs before the leave classes are applied).

## Known latent pattern (out of scope, report-only)

`.hk-list-grow-leave-active` (`HkListTransition.scss`) still combines `position: absolute; width: 100%`; its hosts have definite widths today so it cannot collapse, but any future shrink-to-fit host would hit the same flash.

Version: `@celestia-island/hikari` 0.13.0 → 0.13.1.

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.13.0",
+  "version": "0.13.1",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkBlockingToast.scss
+++ b/packages/vue/src/components/HkBlockingToast.scss
@@ -105,8 +105,11 @@
   transition-property: background-color, border-color, color, box-shadow, opacity, transform;
   transition-duration: var(--hi-duration-normal, 0.3s);
   transition-timing-function: cubic-bezier(0.36, 0, 0.66, 1);
+  /* Only the positioning switch lives here; the leaving slot's box is
+     frozen by the before-leave hook (pinLeaveGeometry), same contract
+     as HkToast — the shrink-to-fit fixed column cannot size leaving
+     cards (a percentage width resolves against the collapsed wrapper). */
   position: absolute;
-  width: 100%;
 }
 
 .hk-blocking-toast-enter-from,
--- a/packages/vue/src/components/HkBlockingToast.test.tsx
+++ b/packages/vue/src/components/HkBlockingToast.test.tsx
@@ -185,4 +185,37 @@ describe("HkBlockingToast", () => {
     await expect(gate).resolves.toBe(true);
     await setLocale("en");
   });
+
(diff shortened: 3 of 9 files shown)


langyo:
Merged with self-hosted runner jobs still queued (environmental, per §7 CI policy): local verification green — vitest 15/15 on the touched suites, full package suite green, vue-tsc --noEmit clean, and the hosted fmt+clippy+test job passed.

langyo:
CI status note: the only non-passing check, `Clippy (stable)`, was **cancelled at the action-download step** (`##[error]A task was canceled.`, job log 98810347013, 0 compile/test steps run) — infra, not code; this PR touches no Rust. All self-hosted jobs sit queued behind an org-wide runner shortage (chest/entelecheia builds from 09:02Z still queued). Will rerun cancelled jobs and merge as soon as a runner frees up.

PR #307, opened 2026-08-28
🐛 Fix popup stacking with kind priority z bands to keep toasts topmost.

langyo:
## Problem

On phones, toasts were painted over by the first-level bottom sheet (every popup is a bottom-up sheet under 768px). Root cause: `usePopupManager` assigned z-index by **registration order only** — the long-lived `HkToast` stack registers once at shell mount and held z=1000, while any modal/drawer opened later got 1002+ and covered the toast.

## Fix

Layering is now decided by **popup kind bands** (`POPUP_Z_BANDS`), not registration order:

| Band | Kind | Notes |
|---|---|---|
| 1000 | modal, drawer | shared overlay band, in-band open-order stacking (drawer over modal still works) |
| 2000 | dropdown | above overlays: select panels/menus opened inside a modal portal to `<body>` and must paint over it (Ant Design tradeoff) |
| 3000 | tooltip | above the surfaces they annotate |
| 4000 | toast | always topmost — a toast can never be buried by a sheet again |

- Within a band: next z = band + (max live slot + 1) × 2; freed slots reclaim automatically; odd slots stay free for each overlay's `+1` content/panel layer.
- Static fallbacks aligned with the scale: `HkToast.scss`/`HkBlockingToast.scss` 9999→4000, `HkTooltip.scss` 10000→3000, `HkModalBreadcrumb.scss` 100001→1500, draggable ghosts 9999→`var(--hk-z-drag, 950)`; `theme.scss :root` tokens mirror `POPUP_Z_BANDS`.
- `POPUP_Z_BANDS` / `POPUP_Z_STEP` exported for hosts; tests rewritten to pin the regression, in-band reclaim, shared modal/drawer band, dropdown-above-overlay (both orders), tooltip<toast.

## Notes for downstream

- Hosts that drag items **inside** an open modal/drawer can raise `--hk-z-drag` (ghosts now sit below the overlay bands by design).
- Keep mounting `<HToast />` before `<HkBlockingToast />` (existing convention; the two containers share the toast band base, DOM order breaks the tie).

## Verification

- Full vitest: 59 files / **490 tests PASS**; `vue-tsc --noEmit` PASS.
- Independent review subagent audit of all 8 popup-manager consumers: SHIP, no stacking regressions.
- Version bump 0.13.0 → 0.14.0.

PR #308, opened 2026-08-28
✨ Rebuild the theme mode group as a plain segmented group and add built-in tab overflow and add buttons.

langyo:
## Summary
Two upstream primitives stop hosts from hand-rolling chrome (P93 wave, user feedback on the theme popup and tab strips):

1. **HkThemeToggle — plain HkSegmented mode group.** The auto-mode solar-altitude strip was an absolutely-positioned overlay button covering the Light/Dark halves, with hardcoded `calc()` geometry mirrored against HkSegmented's touch CSS (a coupling the stylesheet itself flagged). Now: a plain three-segment group (Auto | Light | Dark) with **every segment always enabled** — pressing a side while auto is active *is* the switch to that manual side. The altitude readout becomes a passive caption row **below** the group (auto only, same 60s refresh + memoized geo); the overlay CSS, geometry calcs and resolve-debounce are deleted.
2. **HkTabs — built-in overflow + protruding add button.**
   - `addable` + `addLabel` + `add` emit: a trailing Plus `HkIconButton` composed through `HkHoverRevealAction` (hover-revealed on pointer devices, tap-revealed with linger on touch), protruding at the strip's right edge **outside the scroll viewport** so it never scrolls away.
   - `scrollbar` opt-in: passes the scroller's auto-hiding overlay track through (default off; fades remain).
   - i18n `hikari::tabs.add` in all 11 locales (json + toml); `autoAltitudeTip` copy corrected (no longer claims to be pressable).

Hosts (chest top/bottom tab bars) previously wrapped `HTabs` in their own `HHoverRevealAction` + hand-styled buttons; that entire pattern can now be one `addable` prop.

## Test plan
- [x] `vitest`: full suite **500/500 green** (61 files), incl. rewritten mode-group tests (segments enabled in auto, passive hint only in auto, direct Light switch pins `light`), addable tests (render/emit/labels/outside-viewport assertion) and scrollbar opt-in test
- [x] `vue-tsc --noEmit` clean
- [x] commit-msg lint PASS
- [x] Verification subagent round 1: PASS; minors addressed (tooltip copy, outside-viewport test, mount-time note)


langyo:
CI note: the *Test Coverage* job fails in its **Setup toolchain** step (dtolnay/rust-toolchain date-arithmetic script, exit 1 before any build/test runs). This is a pre-existing repo-wide infra failure: the same workflow fails identically on master (runs at 08:52Z and 10:53Z today) and on prior merged PRs (#306 fix/toast-leave-flash, #307 fix/popup-z-bands). Not caused by this change. Local verification green: full vitest 500/500 (61 files), vue-tsc clean, hosted `fmt + clippy + test (hosted)` PASS, version consistency check PASS locally. Merging under §6.2/§7 recorded-exempt policy; remaining self-hosted jobs are reference per §7.

PR #309, opened 2026-08-29
🐛 Keep the modal breadcrumb visible across menus and stacked sheets.

langyo:
**P93 — the floating window-stack breadcrumb stopped appearing (and then stopped being visible at all).**

Two independent legs, both surfaced by the P91 menu/sheet unification:

1. **Gate blindness** — `HkModalBreadcrumb` only counted `kind === "modal"` registry entries and rendered at >1. Menus, sheets and popouts (`HkSelectPanel`) registered as `dropdown` with no title, so the classic "modal opened from the user menu" was a single modal entry (strip hidden), and sheet-over-sheet stacks were zero entries.
2. **Z burial** — the strip's fallback z (1500) sits *below* the dropdown band (2000+): every open level paints a fixed `rgba(0 0 0 / 40%)` scrim over it. Before the z fallback was lowered, 100001 masked leg 1 whenever the gate passed; now both hit at once.

Fix:
- `HkModalBreadcrumb`: stack segments now count `modal + drawer + dropdown` (tooltips/toasts excluded) — menu → submenu → dialog, sheet → sheet, dialog over dialog all qualify.
- `HkSelectPanel`: registers with `props.title` so layers get real labels ("User menu / Account security / Manage passkeys") instead of "Layer N".
- `--hk-breadcrumb-z` 1500 → 2500: above every popup band's content and scrim (in-band steps of 2 keep real stacks dozens of levels below), still under tooltip (3000) / toast (4000). Mobile already reserves the top band for it (`--hk-sheet-top-inset`).

Gates: vitest 500/500 (61 files).

### DIFF
--- a/packages/vue/src/components/HkModalBreadcrumb.scss
+++ b/packages/vue/src/components/HkModalBreadcrumb.scss
@@ -2,9 +2,10 @@
   position: fixed;
   left: 50%;
   transform: translate(-50%, -50%);
-  /* Above stacked-modal content (overlay band 1000+), below the dropdown
-     band (2000) so open panels are never covered by the strip. */
-  z-index: var(--hk-breadcrumb-z, 1500);
+  /* Above every popup band's scrim (modal 1000+, dropdown 2000+; in-band
+     steps of 2 keep real stacks far below this) so multi-layer windows
+     never bury the strip; still below tooltip (3000) / toast (4000). */
+  z-index: var(--hk-breadcrumb-z, 2500);
   display: flex;
   align-items: center;
   gap: var(--hi-space-4, 0.5rem);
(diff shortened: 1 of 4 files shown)


langyo:
CI 豁免记录（§7/§9.4）：首跑全数 checkout `upload-pack: not our ref`（镜像延迟）；重跑后 Clippy (stable) **真实通过**（5m59s），其余 job 死于 runner 侧 rustup 工具链缓存损坏（`failure removing component 'cargo-x86_64-unknown-linux-gnu'`）与 apt 安装失败——均为环境性。本 PR 仅改 packages/vue 的 .tsx/.scss，本地门禁 vitest 500/500（61 files）全绿。

PR #310, opened 2026-08-29
🐛 Define theme-aware Hk component tokens and make the scroll container scrollbar prop reactive.

langyo:
## Summary
Two chrome-correctness follow-ups from the P93 wave (chest reported segmented controls rendering with washed-out hardcoded light colors inside dark themes):

1. **Theme bridge for the `--hk-*` component tokens.** `HkSegmented` reads `--hk-surface` / `--hk-surface-muted` / `--hk-text` / `--hk-text-muted` / `--hk-primary`, but nothing defined them — every consumer fell through to hardcoded light-theme fallbacks (white surfaces, near-black text, blue focus ring). `admin-tokens.scss` now bridges them to the runtime `--color-*` tokens (same RGB-triplet convention, original values kept as fallbacks; consumer-local definitions still win by load order).
2. **`HkScrollContainer` `scrollbar` is now reactive.** The overlay tracks were built once in `onMounted`; toggling the prop did nothing until remount. The mount path is factored into `mountScrollbars()` and a post-flush watcher tears down and rebuilds the tracks on flips (this is what `HkTabs`' `scrollbar` pass-through forwards to).

## Test plan
- [x] New test: flipping `scrollbar` builds the `.hk-scrollbar-track` elements and removes them on flip-back
- [x] Affected suites (HkScrollContainer / HkSegmented / HkTabs / HkThemeToggle): 30/30
- [x] Full suite 500/500 green + one timing-flaky `HkDateTimePicker` case that passes 3/3 in isolation (unrelated to this change; 50ms stepper timer under full-suite load)
- [x] `vue-tsc --noEmit` clean; commit-msg lint PASS
- [x] npm version 0.15.0 → 0.15.1

PR #311, opened 2026-08-29
✨ Add tablist semantics and a measured tail overlay to the segmented group.

langyo:
## Summary
Makes HkSegmented strong enough to be THE button group for both roles the apps need — a mode picker (radio group) and a compact tab strip — and adds the group-internal info layer the theme toggle's auto mode wants:

1. **`semantics="tablist"`** — same visual primitive as ARIA tabs: `role=tablist`/`role=tab`, `aria-selected`, roving tabindex; for scrollable tab bars with overflow/add affordances HTabs stays the tool.
2. **Arrow-key navigation for both semantics** — arrows/Home/End, wrapping, disabled options skipped, selection follows focus (role=radio/tab buttons get no native arrow behavior; this closes a pre-existing a11y gap for radiogroup mode too).
3. **`overlayFrom` + `#overlay` tail overlay** — a passive info layer covering everything right of option[overlayFrom], measured by the same geometry pass as the sliding thumb (useMeasuredHighlight gains `containerWidth`), carrying the thumb's pill chrome, click-through except for interactive slot content. Pre-measurement it falls back to an equal-share estimate, same cadence as the thumb.
4. **HkThemeToggle auto mode rides the overlay again** — the solar-altitude strip covers the Light/Dark half of the group (autoAltitudeTip restored to explain the press), pressing it resolves to the manual side auto currently renders; the #308 below-group hint row is gone. The strip is now built on the shared group's own capability instead of hand-rolled absolute geometry (`calc((100% - 8px) / 3 + 4px)`).

npm 0.15.1 → 0.16.0.

## Test plan
- [x] tablist roles/aria-selected/roving tabindex; arrows skip disabled + wrap; Home/End; radiogroup also gains arrows
- [x] overlay: absent by default, slot content rendered, equal-share fallback left, measured px geometry (stubbed layout), dropped at last-option anchor
- [x] theme toggle: overlay only in auto, covered segments disabled, press resolves to a manual side (either side accepted — timezone-dependent)
- [x] Full suite 509/509, vue-tsc clean, commit-msg lint PASS

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.15.1",
+  "version": "0.16.0",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkSegmented.scss
+++ b/packages/vue/src/components/HkSegmented.scss
@@ -154,3 +154,45 @@
     transition: none;
   }
 }
+
+/* ── Tail overlay ──────────────────────────────────────────────────────
+ * A passive info layer covering everything right of option[overlayFrom]
+ * (geometry measured by the same pass as the thumb). Carries the thumb's
+ * pill chrome so it reads as part of the group, sits above the segments,
+ * and is click-through — interactive content inside the slot re-enables
+ * pointer events per element. The component always supplies an inline
+ * `left`: an equal-share estimate before measurement (SSR/hidden), the
+ * measured px (plus width) once real layout exists — same cadence as the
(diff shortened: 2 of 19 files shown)


langyo:
CI note: same environmental class as #310 (recorded there) — the self-hosted Rust jobs fail at Setup toolchain (runner-e rustup env `cannot install while Rust is installed`) or at checkout (`upload-pack: not our ref` bare-mirror sync delay). The change is TS/SCSS-only; hosted `fmt + clippy + test` (full compile+test on GitHub runners), lint-commits, validate and version consistency all PASS. Local: full suite 509/509, vue-tsc clean. Merging under §6.2/§7 recorded-exempt; reruns triggered to keep the board honest.

PR #315, opened 2026-08-29
🐛 Align mobile sheet titles with body entries and drop duplicated sheet headings.

langyo:
## Problem (user-reported, mobile webui)

On phones (<768px) every bottom-sheet popup rendered by `HkSelectPanel` (the shared surface behind `HkSelect` and `HkMenu`) had two top-area defects:

1. **Title outdented from the list it names.** The sheet header title sat at a single `--space-16` left inset while body entries' text sits at ~2× that (16px list inset + 16px row padding), so the title visibly hung left of every entry — it read as a stray caption rather than the list's heading.
2. **Duplicated title.** Composition-slot consumers repeat the panel title inside the content: the chest workspace picker passes `title="工作区"` **and** renders its own `s-ws-menu-label` "工作区" section label; chest's BreadcrumbBar all-tags menu passes `title="All tags"` **and** a header slot with the same text. On the desktop popout no header is drawn so the in-content label is the only one — on the mobile sheet both show and it looks like a rendering bug.

## Fix

- `.hk-select-sheet-title` now pads `var(--hk-sheet-title-inset, calc(var(--space-16) * 2))` on the left (trailing inset unchanged). Hosts can tune via the new CSS var. Visual check at 393px width confirms the title lands exactly on the entries' text edge.
- New duplicate-title filter in `HkSelectPanel` sheet mode: while a sheet is open, the **first** non-interactive element whose trimmed text **exactly** equals the panel title is hidden (`.hk-sheet-dup-title`, `display:none !important`). Never hidden: interactive rows (`button/a/input/...`), select option blocks (`.hk-select-option`), menu rows (`.hk-menu-row`), and wrappers whose text only matches because they contain such rows (`querySelector` guard — a data-driven `HkMenu title="X" items=[{label:"X"}]` keeps its body). Re-synced while open by a `MutationObserver` (childList + characterData + class attributes) plus a `props.title` watcher, so async-loaded lists neither miss a late duplicate nor keep a stale one hidden. Close/unmount disconnect the observer without restoring the class (restoring mid leave-transition would flash the duplicate back in during the slide-down). A same-tick open→close can't arm the observer on the leaving panel, and a `sheetOnMobile` prop flip mid-open tears the filter down with the sheet branch.
- The desktop popout path is untouched (it draws no header, so in-content headings stay there); `HkDrawer` / `HkModal` sheets are not affected (their titles already align with their 16px body insets).

## Consumer impact

shittim-chest compiles hikari from source (vite alias, `workspace:*`), so the chest workspace picker and all-tags sheet are fixed by this PR alone — no chest code change needed. Version bump 0.19.0 → 0.19.1 in-PR per policy.

## Verification

- New tests: 7 `HkSelectPanel` cases (hide exact dup; partial/non-match kept; interactive rows + wrappers + option blocks never hidden; desktop popout untouched; late-added dup hidden via observer; retitle restores).
- Full suite: **59 files / 510 tests green**; `vue-tsc --noEmit` exit 0.
- 3-round independent subagent verification cycle: all PASS (observer convergence, lifecycle ordering, i18n exact-match across 11 locales, chest consumer audit, repo hygiene — no secrets, no CHANGELOG, 4-file diff).
- Visual: static harness at 393px — BEFORE title at 16px with duplicated label; AFTER title on the 32px entry-text edge with the duplicate filtered (group labels with different names preserved).

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.19.0",
+  "version": "0.19.1",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkSelect.scss
+++ b/packages/vue/src/components/HkSelect.scss
@@ -214,7 +214,14 @@
 
 .hk-select-sheet-title {
   flex-shrink: 0;
-  padding: 0 var(--space-16) var(--space-8);
+  /* Left inset tracks the sheet ENTRIES' text edge, not their card
+   * edge: the list pads option blocks in by --space-16 and rows carry
+   * roughly another --space-16 of leading padding inside, so the old
+   * single --space-16 left the title hanging left of every row it
+   * named instead of reading as the list's heading. Hosts tune it via
+   * --hk-sheet-title-inset; the trailing inset stays --space-16. */
+  padding: 0 var(--space-16) var(--space-8)
+    var(--hk-sheet-title-inset, calc(var(--space-16) * 2));
   font-size: var(--text-xs);
   font-weight: 600;
   letter-spacing: 0.02em;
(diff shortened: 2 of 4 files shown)


langyo:
Merge gate note: lint-commits + hosted fmt/clippy/test + validate are green; local verification green (vitest 59 files / 510 tests, vue-tsc clean). Remaining self-hosted jobs (Clippy nightly/stable, Test nightly/stable, Test Coverage, version consistency, Lint, Run tests) are still queued on the shared runner pool past the ~15 min window — recorded-exempt per PLAN CI policy (TS/SCSS-only diff; the Rust suites are unaffected). Stale queued jobs will be cancelled after merge.

PR #316, opened 2026-08-29
🐛 Keep the fixed token usage pill at least as wide as its content.

langyo:
## Problem

The cruise "nodes" board in shittim-chest showed a native horizontal scrollbar (with OS arrow buttons) inside every lane label card, below the model row.

Root cause: the `data-fixed` variant of `HkTokenUsageBadge` capped the pill at a hard `width: 100px`, but two 6-character values (`197.2k` / `38.9k`) plus the arrows and the inter-stat gap render ≈112px. The pill's box stayed at 100px (consumers set `flex-shrink: 0`), so its content spilled ~12px invisibly past the box — and that spill extends the scrollable overflow area of every ancestor. The lane label's model-lines list sets `overflow-y: auto`, which makes `overflow-x` compute to `auto` as well, so the 12px spill rendered a real horizontal scrollbar.

## Fix

- `[data-fixed]`: `width: 100px` → `min-width: 100px; width: max-content`. Short values keep the 100px floor (the `space-between` spread and row alignment across sibling lines stay put — the anti-jitter intent), longer values hug their content so nothing spills.
- Pin the contract with a new `HkTokenUsageBadge.test.tsx` (render tests + a source-contract test forbidding a hard 100px cap).
- Version bump 0.19.0 → 0.19.1.

## Verification

- `vitest run src/components/HkTokenUsageBadge.test.tsx`: 3/3 green.
- Full hikari suite: 506/506 green.
- DOM measurement rig against the deployed CSS confirmed: before the fix `scrollWidth` exceeded `clientWidth` by 12px in every width case; after the fix scrollWidth == clientWidth and the culprit list is empty.
- Consumer audit (round-1 + round-2 verification subagents): the only `fixed` consumers are shittim-chest's CruiseDock (no scroll containers, ~318px inner) and MindMapView lane label / mobile group card (identical shrink chains); TodoLogModal and the node-card stats row never pass `fixed`. Worst-case line at the 232px card floor ≈183px ≤ 198px inner — the model tag ellipsizes first, nothing visible clips.

Downstream: shittim-chest adds an explicit `overflow-x: hidden` on the model-lines list and will pair with this version (its CI clones hikari master fresh).

### DIFF
--- a/packages/vue/src/components/HkTokenUsageBadge.scss
+++ b/packages/vue/src/components/HkTokenUsageBadge.scss
@@ -7,7 +7,18 @@
   color: rgb(var(--color-muted));
 
   &[data-fixed] {
-    width: 100px;
+    /* Stable-width mode: short values keep the 100px floor (the
+       `justify-content: space-between` spread and row alignment across
+       sibling lines stay put), but the box must be allowed to hug the
+       content when the rendered digits need more room — a hard 100px
+       cap is narrower than two 6-character values ("197.2k" / "38.9k"
+       plus arrows and the inter-stat gap ≈ 112px), and the invisible
+       spill extends every scrollable ancestor's scrollWidth. In the
+       shittim-chest cruise lane label that grew a native horizontal
+       scrollbar under the model row (`overflow-y: auto` implies
+       `overflow-x: auto`). */
+    min-width: 100px;
+    width: max-content;
     justify-content: space-between;
   }
 }
--- a/packages/vue/src/components/HkTokenUsageBadge.test.tsx
+++ b/packages/vue/src/components/HkTokenUsageBadge.test.tsx
@@ -1,0 +1,59 @@
+import { afterEach, describe, expect, it } from "vitest";
+import { createApp, h } from "vue";
+import { readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+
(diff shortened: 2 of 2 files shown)


langyo:
Merge exemption record (§6.2 / CI policy): the required `lint-commits` job has been stuck in `queued` on the shared self-hosted runner for ~50 minutes (multiple other queued Commit Message Lint runs ahead of it — infra queue saturation, no failure). Local verification is fully green: single-file vitest 3/3, full hikari suite 506/506, vue-tsc 0 errors. The commit subject follows the enforced format (gitmoji + one English sentence, no colon prefix — same shape as every previously merged PR that passed this linter), and the `validate` required check passed. Merging per the environment-only-failure exemption; the queued lint run is harmless and may be cancelled afterwards to free the runner.

langyo:
Update on the earlier exemption record: the queued `lint-commits` run eventually executed but failed with `fatal: Invalid revision range 89c7825..c109636` (exit 128) — the range spans the pre-rebase SHA that no longer exists after the branch was rebased onto master (#315 landed mid-flight), so the linter crashed on git range resolution rather than rejecting any message. Local check with the same linter (`celestia-devtools commit-msg-lint check`) passes with exit 0 on the actual subject. This is an infra/tooling artifact per §6.2, local verification remains green (vitest 506/506, vue-tsc 0 errors), so merging.

langyo:
Superseded by #318 (fresh branch so the lint range resolves after the rebase).

PR #289, opened 2026-08-26
✨ Make the select panel one back-gesture layer that closes before the page navigates.

langyo:
## What

User-reported defect: the dropdown panel opened via `HkSelectPanel` (the shared surface behind BOTH the real `HkSelect` dropdown and custom invocations like the chest cruise clock popover) did not count as a back-gesture layer — pressing browser/system back while it was open navigated the page/window instead of first closing the panel.

Fix: integrate the shared back-guard service (`runtime/backStack.ts`) — the same window-first back priority `HkModal`/`HkMenu` already use:

- `createBackGuard({ onBack: close })` at setup
- open pushes ONE marked history entry (release-then-push form, the HkMenu normalizer pattern — a same-tick close→reopen keeps the reopened panel guarded; the service's N1 semantics at backStack.ts:279-285)
- any ordinary close (row click, Escape, outside click, scrim, breakpoint cross) rewinds the entry — no dead back left behind
- unmount rewinds synchronously — no stranded marker

Single implementation, every consumer inherits: real dropdown, custom invocation, desktop popout, mobile sheet. No new prop (always on, like HkMenu mobile sheets).

## Tests (12 in HkSelectPanel.test.tsx, +4 back-guard)

- back closes the open panel, not the page (marker asserted, then history.back() → closed + clean base)
- outside-click close rewinds the entry (hard asserts after each wait)
- unmount-while-open rewinds (no stranded marker)
- same-tick close→reopen keeps the reopened panel guarded (with the flush actually crossing the close branch — mutation-verified: removing close-branch `release()` fails both rewind tests)

## Verification

3-round cycle (count restarted once when a round caught a vacuous test — twice, both fixed with real close paths + hard asserts + mutation checks). Local: vue-tsc 0 errors, vitest 49 files / 407 tests green (neighbors: backStack 13, HkModal 5, HkMenu 11, HkSelect 2). Version 0.6.4 → 0.6.5.

PR #319, opened 2026-08-29
✨ Let a menu leaf select abandon the back-guard rewind so its navigation survives.

langyo:
## Root cause

Clicking a leaf row in an HkMenu popup/sheet (e.g. the user menu's **Admin Panel**, Log out, or the tag menu in BreadcrumbBar) closes the panel and starts the consumer's action in the same tick. Since #289 the panel is a back-gesture layer: opening pushed a marked history entry, and closing queues a deferred rewind whose flush re-validates the history top **on the next macrotask**.

An async `router.push` from the select handler commits its pushState only **after** that flush, so the flush still saw its own marker on top and issued `go(-n)` — bouncing the page back onto the closed menu's entry and silently discarding the navigation. The admin panel (and every other menu leaf that navigates) never opened, on every deployment carrying #289+ (dev + demo both affected; direct URL entry to /backend worked, which masked the break).

## Fix

- `backStack.abandon()`: retract the queued rewind claim and snap the counters **without touching history** — pushed markers stay where they lie as inert dead markers, released in place by the existing popstate cleanup.
- `HkSelectPanel` exposes `abandonBackGuard()`; `HkMenu.onItem` abandons every panel's rewind (desktop popouts + mobile sheets, all cascade levels) before emitting select + closeAll on a leaf click.
- Branch clicks, plain close paths (overlay/Escape/collapse), HkSelect and HkPopupSelect selection flows keep the `release()` rewind — untouched.

Trade-off: a leaf select that does NOT navigate (locale/theme switch) leaves one inert marker — the first back press consumes it silently with no URL change (the module's documented dead-marker contract), the next back works normally. Every leaf select that DOES navigate now works; before, it never did.

## Verification

- vitest 59 files / 514 tests green (3 new abandon unit tests, rewritten mobile leaf-select test, new desktop navigation-bounce regression test)
- vue-tsc --noEmit clean
- Independent verifier subagent: PASS, no blockers; adversarial traces (back-after-leaf, destroy-after-abandon, abandon-then-reopen) all confirmed safe

npm track 0.19.3.

### DIFF
--- a/packages/vue/src/components/HkMenu.test.ts
+++ b/packages/vue/src/components/HkMenu.test.ts
@@ -158,6 +158,11 @@ afterEach(async () => {
   while (mounts.length) mounts.pop()?.unmount();
   // Unmounting restores history asynchronously; wait for it to land so a
   // pending traversal cannot pop the NEXT test's freshly pushed entries.
+  // Leaf-select tests ABANDON their rewind on purpose, so the marker
+  // entry stays current as an inert dead marker — de-mark it in place
+  // (the same way a landing popstate would) or the until() below would
+  // spin to its timeout on an intentionally kept marker.
+  window.history.replaceState(null, "");
   await until(() => window.history.state?.__hkBack === undefined);
   // happy-dom never fires transitionend, so leave transitions never
   // finish and teleported panels linger on <body> — strip them or the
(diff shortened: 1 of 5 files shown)


langyo:
Merge gate note (§6.2/§7 exemption): the hosted code-level checks (`fmt + clippy + test (hosted)`, `validate`) are green, local verification is fully green (vitest 59 files / 514 tests, vue-tsc --noEmit clean, commit-msg hook passed), and two independent verifier subagents returned PASS with no blockers. All self-hosted jobs have been queued-only for >20 minutes at ~01:47 (runner queue, environment-only); merging now per the CI usage policy. Any later self-hosted failure that turns out code-level will be fixed in a follow-up before the downstream chest rebuild consumes this.

PR #320, opened 2026-08-30
🐛 Tighten mobile sheet spacing and bottom-anchor autoFollow content.

langyo:
## User feedback (P93g, 2026-08-29)

Mobile popups with bottom control buttons docked too far off the screen bottom, and the node chat-history thinking-records list left too much distance from the bottom in auto (follow) mode.

## Changes

- **Sheet spacing**: `--hk-sheet-bottom-gap` default 0.5rem → 0.375rem; footer padding base `calc(0.5rem + var(--hk-sheet-footer-gap, 0.75rem) + env(safe-area-inset-bottom))` → `calc(0.375rem + var(--hk-sheet-footer-gap, 0.5rem) + env(...))`. Both hooks stay host-tunable; the safe-area inset still stacks on top so home-bar devices never lose tappable chrome.
- **autoFollow content bottom-anchors while it fits** (`.hk-scroll-container-autofollow-content`: flex column + justify-end + min-height:100%): the agent thinking-step stream hugs the panel bottom in auto mode instead of top-hanging with a dead band; once content overflows the wrapper grows and ordinary scroll + follow resumes (inert flex-end).
- Contract test pins the new default. vitest **59/510**, vue-tsc clean, 3-round subagent verify PASS. Version **0.19.3**.

Paired chest consumer wave follows.

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.19.2",
+  "version": "0.19.3",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkModal.scss
+++ b/packages/vue/src/components/HkModal.scss
@@ -318,7 +318,7 @@
     top: var(--hk-sheet-top-inset, 4.25rem);
     // --hk-sheet-bottom-gap keeps the docked sheet from touching the very
     // bottom edge (host-tunable; chest narrows the default visual gap).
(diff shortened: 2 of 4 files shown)


langyo:
Merging per §6.2: required check lint-commits PASS (plus validate, Verify version consistency, hosted fmt+clippy+test all PASS). The 9 self-hosted jobs failed at Checkout with the documented §9.4 bare-mirror sync delay (`git upload-pack: not our ref` — mirror hadn't picked up the new ref; self-heals on rerun). Change is packages/vue-only; local gates: vitest 59/510, vue-tsc clean, 3-round subagent verify PASS.

PR #325, opened 2026-08-30
✨ Add mobile viewport policy hook, clear tap highlights and drop the last SFC.

langyo:
## Summary
Upstream half of the 2026-08-30 mobile web UX wave (user-reported: blue tap-highlight overlay on buttons + phones refusing zoom across all webuis).

- `foundation.scss`: html-level `-webkit-tap-highlight-color: transparent` (inherited) kills the default blue tap overlay; hikari's own :active/:hover states stay the only tap feedback.
- `runtime/mobileViewport.ts`: `applyViewportPolicy()` — idempotent bootstrap hook. Normalizes the viewport meta so mobile browsers never refuse pinch zoom: strips `user-scalable=no` / `maximum-scale` caps / blocking `minimum-scale`, writes explicit `user-scalable=yes`, preserves `initial-scale` / `viewport-fit` / vendor keys; `allowZoomOut` pins `minimum-scale` (default 0.25) so desktop-first layouts can zoom out to fit on iOS (which clamps zoom-out at minimum-scale).
- 13 vitest cases (`mobileViewport.test.ts`); exported from runtime + package index.
- Demo: call the hook at bootstrap, add the missing viewport meta, convert `DemoApp.vue` → `DemoApp.tsx` + `demo.scss` — **last SFC in the tree removed** (house rule: tsx + scss only).
- Version 0.19.6 → 0.19.7 (ordered after #323 / the input-affix wave per PLAN §0 claim).

## Verification
- `vue-tsc --noEmit` clean on the package.
- `vitest run src/runtime/mobileViewport.test.ts` — 13/13 green.
- No consumer API breakage: purely additive exports + one demo-file rename.

Downstream adoption (chest, erp.celestia.world, easy-hydro-erp, e.celestia.world, noa, arcaea, sysl, github.io, evernight) follows in per-repo PRs calling `applyViewportPolicy()` at bootstrap.

PR #326, opened 2026-08-30
🐛 Let menu panels size to the viewport and clamp flipped panels on-screen.

langyo:
## What

The desktop dropdown popout (`.hk-select-popout`) was hard-capped at `max-height: 240px` — right for small select lists, but it clips richer panels:

- **User account menu** (HkMenu levels render inside HkSelectPanel popouts): identity header + menu rows ≈ 320px+ → a pointless scrollbar even on tall viewports (user screenshot feedback 2026-08-30).
- **Platform switcher** (chest, custom content inside HkSelectPanel): tail rows scrolled away.

## Fix

1. **`HkSelect.scss`** — the cap is now viewport-relative with the old 240px as floor:
   ```scss
   max-height: max(240px, min(36rem, calc(100vh - 32px)));
   max-height: max(240px, min(36rem, calc(100dvh - 32px)));
   ```
   vh line first as the fallback for engines without `dvh` — same two-declaration convention as HkModal's sheet styles. The mobile bottom-sheet sizing is untouched.
2. **`HkSelectPanel.tsx`** — the taller ceiling exposed a latent gap in `positionPanel()`: the single bottom→top flip never re-checked the result, and the final `top` was applied verbatim, so a mid-viewport anchor with a near-max panel could flip into a **negative top** (panel header off-screen). The final top is now clamped into `[VIEWPORT_PAD, innerHeight − VIEWPORT_PAD − ph]`; internal popout scroll takes over when content exceeds the CSS cap.
3. Version 0.19.6 → 0.19.8 (rebased over #325, which took 0.19.7).

## Verification

- `vue-tsc --noEmit`: clean.
- `vitest run`: HkSelectPanel 20/20 (incl. new negative-top clamp test — verified it fails with `expected -200 to be ≥ 8` when the clamp is disabled), HkMenu 24/24, HkSelect/HkSelectSheet/HkModal.sheetgap all green (56/56 across the related suites).
- sass compiles both declarations verbatim; the dvh-fallback ordering convention is enforced by the existing HkModal.sheetgap test.
- Independent cross-review passed (CSS math validity, consumer sweep, version-sync check).

CI is reference-only per workspace policy; local verification above is the gate.

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.19.7",
+  "version": "0.19.8",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkSelect.scss
+++ b/packages/vue/src/components/HkSelect.scss
@@ -94,7 +94,12 @@
   position: fixed;
   /* z-index comes from the popup manager (inline) — registered overlay. */
   min-width: 180px;
-  max-height: 240px;
+  /* Menu/switcher panels overflow a bare 240px cap even on tall viewports
+     (user feedback 2026-08-30: identity header + rows ≈ 320px+). Keep 240px
+     as the floor for select lists on short viewports, otherwise let the
+     panel size to its content up to a viewport-relative ceiling. */
+  max-height: max(240px, min(36rem, calc(100vh - 32px)));
+  max-height: max(240px, min(36rem, calc(100dvh - 32px)));
   display: flex;
   flex-direction: column;
   gap: var(--space-2);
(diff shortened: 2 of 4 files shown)


langyo:
Merging per workspace CI policy (§6.2): all self-hosted jobs sat in the runner queue >12 minutes (concurrent waves saturated the pool) — environment-only. Gates green locally and on hosted CI: `CI (GitHub-hosted fallback)` ✓, `PR Title Check` ✓, `validate` ✓; local `vue-tsc --noEmit` clean and 56/56 related vitest tests green (incl. the new clamp test, verified failing without the fix). Commit subjects were validated by the org commit-msg hook at commit time.

langyo:
Post-rebase note: branch rebased over #325 (which took 0.19.7), version here is now 0.19.8. Re-verified after rebase: vue-tsc clean, HkSelectPanel+HkMenu 44/44 green. Remaining red checks are the documented self-hosted toolchain flake (llvm-tools-preview component conflict, §9.4 class) and queue timeouts — environment-only; the hosted fallback, stable Clippy/Test, lint-commits and version-consistency are all green.

PR #324, opened 2026-08-30
🐛 Pin input suffixes right, float dropdown sheets and align theme rows with the dropdown grammar.

langyo:
## Problem (user-reported, mobile webui, three screenshots)

1. **Localized input chip on the wrong side, placeholder underneath it.** The language chip of `HkLocalizedInput` (rendered in `HkInput`'s suffix slot) sat at the **left** edge of the field with the scrolling placeholder sliding underneath it. Root cause: the input element is absolutely positioned over the whole box, the affixes are the only in-flow flex children, so a lone suffix parked at flex start; and nothing reserved the chip's width for the text line or the marquee window.
2. **Dropdown sheets don't read as the standard sheet family.** `HkSelectPanel`'s mobile sheet (the shared surface behind `HkSelect` / `HkMenu` / `HkPopupSelect` — including the localized input's language menu) was glued to `bottom: 0`, while `HkModal`'s mobile docking floats on `--hk-sheet-bottom-gap`. Same for `HkPopover`'s `sheetOnMobile` mode (theme toggle) — hard `bottom: "0"` inline.
3. **Theme sheet options don't match the standard.** `HkThemeToggle`'s rows used bespoke 6×8 tight styling (small text, borderless 12% wash), visually unrelated to the `.hk-select-option` grammar every dropdown row speaks.

## Fix

- **HkInput**: `.hk-input-affix.hk-input-suffix { margin-inline-start: auto }` pins the trailing affix to the right edge. Both affixes are measured (ResizeObserver-backed stable ref callbacks) and published as `--hk-input-affix-start-w` / `--hk-input-affix-end-w` on the box; the input element's horizontal padding and a new `.hk-input-box .hk-placeholder-marquee` inset consume them — typed text, caret, native placeholder AND the scrolling marquee always end short of the chip. Affix removal is judged after the patch settles (Vue fires function refs with null **before** detaching the node — a synchronous `isConnected` guard was proven dead code in review and pinned by test). Textarea variant keeps its plain padding (in-flow, unaffected).
[...]

PR #328, opened 2026-08-31
🐛 Add breathing room between the step flow header and its body.

langyo:
## Summary
- HkStepFlow rendered its step-indicator timeline flush against the step body (zero gap), which made modal wizards (e.g. shittim-chest 新建视图/新建标签 wizards, provider wizard) read as cramped — the indicator row nearly touched the first card below it.
- Add a default 16px (`--space-16`) gap between `.hk-step-flow > .hk-timeline` and the body, tunable per host via a new `--hk-stepflow-header-gap` custom property.
- Sticky-header mode folds the gap into the pinned header's `padding-bottom` (replacing the old hardcoded 8px) so the surface tint stays continuous while the body scrolls underneath; `margin-bottom` is zeroed there to avoid double spacing.

## Verification
- `sass` compile of the modified SCSS: clean.
- vitest: HkStepFlow + HkTimeline suites 25/25 green.
- Browser DOM check on a minimal harness with the compiled CSS: non-sticky gap = 16px; sticky mode margin 0 / padding-bottom 16px as designed.
- Project-wide scan: shittim-chest (3 consumers) needs no consumer-side change; arona / plana / e.celestia.world / entelecheia have zero step-flow usage; easy-hydro-erp vendors only HkTimeline (standalone usages, unaffected by the `.hk-step-flow`-scoped selector).

Version bump: packages/vue 0.19.9 → 0.19.10.

PR #329, opened 2026-08-31
🐛 Show locale codes only in menus and unify popup scrollbars.

langyo:
## Part A — HkLocalizedInput locale chip
The field chip now shows ONLY the language label (简体中文); the parenthesized code appears ONLY in the opened menu rows (简体中文 (zh-Hans)) — switch rows, Add-language and Delete-language cascade children included. Chip stays pinned to the input's right edge (suffix pin from #324). User-reported via the chest add-panel wizard screenshot.

## Part B — one overlay scrollbar system
- New composable `useOverlayScrollbar.ts` (attachOverlayScrollbars) extracts HkScrollContainer's track/thumb engine so every scroll region renders identical overlay chrome; explicit `host` option for teleported popups; reference-counted positioning guard.
- Shared chrome CSS moved to theme `_scrollbar.scss`; dead `.hk-obs-*` legacy system deleted.
- **Zero `!important` added (library policy)**; the old HkPopupSelect `!important` re-enable hack is deleted outright; the pre-existing `::-webkit-scrollbar` thin-bar blocks (HkSelect popout/sheet, HkModal, HkDrawer, HkSidebar) are replaced by plain-declaration native-bar hiding + the overlay.
- Attached across: HkSelectPanel (popout + sheet), HkModal, HkDrawer, HkSidebar, HkPopover, HkKeywordSearchModal, HkToolBlock (3 panes), HkTable (h), HkMarkdownRenderer code blocks (h), HkProgressDialog, HkProtocolModal. Each viewport is wrapped/anchored by an EXACT host box so rails hug their own region only.

## Behavior notes for consumers
- `HkSelectPanel.expose.panelEl()` now returns the fixed host wrapper on desktop (panel + tracks): dismissal containment and row scoping both verified unaffected; track/thumb interaction no longer dismisses the open popup (regression test pinned).
- `matchAnchorWidth` keeps matching anchors narrower than 180px (180px floor moved onto the host element so the inline min-width wins).
- Faded rails are click-through; they become interactive only in the visible flash/hover window.
- HkTextarea deliberately keeps the native bar (editable control). HkLogWindow was already HkScrollContainer-based.
- Version 0.19.9 → 0.19.10. shittim-chest picks this up automatically via the sibling alias on its next build.

## Verification
- 3-round adversarial verify cycle (R1: track-click dismissal blocker + min-width major; R2: mispositioned-rail major + pointer-events minor; R3: SHIP).
- vitest 65 files / 560 tests green · vue-tsc --noEmit clean · `diff | grep '^+' | grep '!important'` = 0 · foundation SCSS compile OK.
- Known pre-existing (not this branch): sass CLI failure of the packages/components chain on master; HkSelect.scss:278 / HkModal.scss:304 legacy !importants documented for a follow-up purge wave.

### DIFF
--- a/packages/theme/styles/_scrollbar.scss
+++ b/packages/theme/styles/_scrollbar.scss
@@ -1,93 +1,72 @@
-// _scrollbar.scss — Custom overlay scrollbar CSS.
-// Adopted from shittim-chest's SScrollContainer overlay system.
-// This hides native scrollbars and provides CSS for the overlay track/thumb
-// that a JS component (or lagrange's vanilla JS) creates imperatively.
+// _scrollbar.scss — the shared overlay scrollbar chrome.
+//
+// ONE scrollbar system for the whole library: the `.hk-scrollbar-track` /
+// `.hk-scrollbar-thumb` DOM is created imperatively by
+// packages/vue/src/composables/useOverlayScrollbar.ts
+// (attachOverlayScrollbars) and styled here. HkScrollContainer keeps its
+// container-specific rules in HkScrollContainer.scss; every other scroll
+// region in the library attaches the same overlay and hides its native
+// bar with the standard two-liner:
+//
+//   scrollbar-width: none;
+//   &::-webkit-scrollbar { display: none; }
 
-// Hide native scrollbar on any element with .hk-scroll-container
-.hk-scroll-container {
-  overflow: auto;
-  scrollbar-width: none !important;
-
-  &::-webkit-scrollbar {
-    display: none !important;
-  }
-}
-
-// Overlay scrollbar track (vertical)
-.hk-obs-track {
+.hk-scrollbar-track {
   position: absolute;
-  top: var(--hi-scroll-inset, 4px);
-  bottom: var(--hi-scroll-inset, 4px);
-  right: var(--hi-scroll-inset, 4px);
-  width: var(--hi-scroll-size, 8px);
+  z-index: 10;
+  background: transparent;
+  /* Click-through while faded: an invisible rail must never steal
+     clicks from the content beneath it (the track hugs the region's
+     edge, exactly where end-of-line links/buttons live). The rail
+     becomes interactive only while it is visible — the flash window
+     after a scroll (and while hovered/dragged within that window). */
   pointer-events: none;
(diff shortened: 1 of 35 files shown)


langyo:
CI environment exemption (§6.2/§9.4): the 4 failing jobs (Clippy nightly, Format, Test nightly, Test Coverage) all died at the `actions/checkout` step with `git upload-pack: not our ref` — the self-hosted bare-mirror ref-sync delay, before any step ran. This PR contains zero Rust changes (Vue/TS/SCSS only); the hosted `fmt + clippy + test` job PASSED, covering the same checks on a fresh checkout. Local verification is fully green: vitest 65 files / 560 tests, vue-tsc --noEmit clean, lint-commits + version-consistency pass, foundation SCSS compile OK, diff adds zero `!important`.

langyo:
Follow-up on the environment exemptions: the reruns confirm the pattern — self-hosted nightly-toolchain jobs (Format/Clippy-nightly/Test-nightly/Test-Coverage) fail in SETUP (rustup install / checkout ref-sync), never reaching the actual steps, and the ci.yml `Run tests` job hung 10+ min in the Post-Restore-Rust-Cache step (known shared-cache tar contention) and was cancelled per §7.3. For this Vue/TS/SCSS-only PR the authoritative equivalents are all green: hosted fmt+clippy+test PASS, Test (stable) PASS, lint-commits PASS, version-consistency PASS, plus local vitest 65/560 + vue-tsc clean.

langyo:
Third environment note (§6.2/§7.1): after two upstream merges (#328, #330) forced two rebase+force-push cycles, the `Lint all new commits on push` step failed with `Invalid revision range <old-head>..<new-head>` — the runner's reused workdir still held the force-pushed-away commit object, defeating the workflow's own unreachable-object guard (infra artifact, not a message violation). The identical subject `🐛 Show locale codes only in menus and unify popup scrollbars.` PASSED lint-commits on both prior heads (b427ec1: 49s pass, 6fc79e2: 3m5s pass), and the PR-title step (same subject, squash source) passed in every run. Merging under the §6.2 exemption with local verification fully green (vitest 65 files / 560 tests, vue-tsc clean, foundation SCSS compile).

PR #333, opened 2026-08-31
⚡ Cancel superseded workflow runs via concurrency groups.

langyo:
Adds a `concurrency` group (`workflow name + PR number / ref`) with `cancel-in-progress: true` to the CI workflows so a newer run on the same PR or branch automatically cancels the queued/running predecessor.

Motivation: org runner data (2026-08-28..31) shows 226/495 runs in this repo started after a newer run on the same branch already existed, wasting self-hosted runner slots. Complements the earlier pull_request trigger-type retrofit.

Notes:
- Groups are keyed by workflow name (unique per repo), so ci.yml and ci-hosted.yml keep independent groups.
- workflow_dispatch runs on a branch join that branch's group; a manual dispatch may cancel a running push run on the same branch (intended).
- Release/publish/docs workflows are intentionally untouched.

### DIFF
--- a/.github/workflows/ci-hosted.yml
+++ b/.github/workflows/ci-hosted.yml
@@ -7,6 +7,10 @@ on:
   pull_request:
     types: [opened, reopened, ready_for_review]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 permissions:
   contents: read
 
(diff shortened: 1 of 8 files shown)


langyo:
Merge record (§6.2 environmental exemption): checks are queued on the shared self-hosted runner queue, not failing. This is a YAML-only change (adds top-level `concurrency` groups to 8 CI workflows); verified locally in 3 rounds — (1) all 8 files parse as YAML with a top-level `concurrency` key via PyYAML, (2) `git diff --stat` shows exactly the 8 expected files at +4 lines each and nothing else, (3) `git diff | grep -iE 'password|secret|token|api_key'` returns nothing. Commit passed the local commit-msg hook lint; noa PR title check already SUCCESS. Merging per workspace CI policy (CI is reference, not a gate) without waiting on the queue.

PR #336, opened 2026-08-31
✨ Give the theme clock a real geolocation feed and a provider hook.

langyo:
## Summary

The theme engine's `system` mode computed the solar period **exactly once, at module init, from `DEFAULT_GEO_LOCATION`** — auto mode was pinned to Shanghai's daylight at load time for every consumer, however far from UTC+8, and never flipped as dusk passed while the page stayed open. `HkThemeToggle` fetched a real fix, but only for its altitude readout; the engine never saw it.

## What this does

**Geolocation cascade (`useSolarTime`)** — `getGeolocation()` now resolves through, in order:

1. a host-registered provider — `setGeolocationProvider(fn)`: native shells (Tauri 2 geolocation plugin, Electron IPC, a saved user preference) feed the real device fix; return `null`/throw to fall through;
2. standard `navigator.geolocation` — **only when `navigator.permissions` already reports `granted`**; a theme must never be the thing that pops the browser's location prompt;
3. the existing IP lookup (timeout-guarded);
4. the synchronous timezone-offset estimate.

Successful fixes are memoized per page (clock, toggle readout and wallpaper clocks share one lookup); re-registering a provider resets the memo; the timezone estimate is **not** memoized, so an offline boot heals on the next `getGeolocation()` / `refreshThemeClock()`.

**Theme clock (`useTheme`)** — `initTheme()` starts a clock: one async geo resolve plus a visibility-aware 5-minute tick (`scheduleInterval`) that recomputes the period from the cached coordinates and re-applies the theme in system mode on an actual day↔night flip. First paint uses the timezone estimate instead of the Shanghai constant. `refreshThemeClock()` / `stopThemeClock()` are exported for late provider registration and host teardown.

**API surface (additive)** — `useTheme()` additionally returns read-only `geo` and `period`; `HkThemeToggle`'s altitude readout now consumes them instead of its own module-level fetch-once hack, so the readout, the toggle glyph and the engine always agree. Existing exports/signatures are unchanged — downstream apps (plana, arona, chest, wowsp) pick the fix up with zero code changes; native-shell apps can opt into the provider hook.

## Tests

10 new tests: cascade ordering (provider win / `null` / throw, granted-vs-`prompt` browser geo without prompting, unpinned estimate), memoization + re-registration reset, clock lifecycle (init starts it, late provider honored via `refreshThemeClock`, stop/restart). Full suite: 67 files / 565 tests green; `vue-tsc` clean.

🤖 Generated with [ZCode](https://zcode.ai)


langyo:
Rebasing onto the updated master (it grew #333–#339 while the runner queue was starved); continuing in a fresh PR so the commit-lint push-event range stays valid.

PR #330, opened 2026-08-31
✨ Merge dynamic option runs into the tab strip and complete button-group motion.

langyo:
## What

The theme-mode segmented control framed its auto-mode solar-altitude strip with an unwanted box (the retired `overlayFrom`/`#overlay` layer carried its own border + opaque surface), and button-group click motion was only half wired. This PR gives HTabs first-class **dynamic merged-option rendering** and completes the motion system.

## Changes

- **HTabs**: new `mergeKeys` prop + `#merged` slot — a contiguous run of options collapses into ONE merged cell (`.hk-tabs-merged`), a real flex child of the track with naked unselected-option chrome. `overlayFrom`/`#overlay`/`.hk-tabs-overlay` are removed (0.20.0 breaking).
- **Merged cell joins the animation context**: the sliding indicator measures it when the active key is merged; option/cell inserts and removals cross-fade via the tagless `hk-tabs-swap` TransitionGroup.
- **Geometry-pinned leave ghosts**: `pinLeaveGeometry` gains `anchorX`/`box` options; HkTabs snapshots child boxes in `onBeforeUpdate` (Vue 3.5 lifts the first leaving sibling to `position:absolute` synchronously in the same patch pass — live reads would freeze reflowed positions), and cancelled leaves drop the pins. Specificity-safe ghost lift via `.hk-tabs-list > .hk-tabs-swap-leave-active`.
- **A11y/robustness**: keyboard nav + roving tabindex skip merged options and the group stays reachable when the active option is merged; whole-strip `disabled` mutes the cell; non-contiguous or slot-less `mergeKeys` degrade to plain triggers.
- **All button groups complete the click motion**: indicator fades in on first measurement (no pop-in), triggers give tactile `:active` scale feedback, dynamic option changes cross-fade.
- **HkThemeToggle**: the altitude strip fills the merged cell (full-bleed press, pointer cursor, press feedback); track height stays trigger-driven across auto↔manual; hoisted `AUTO_MERGE_KEYS` kills redundant watcher fires.
- **Tests**: 552/552 green incl. new merge-cell, contiguity-degradation, keyboard-skip, indicator-framing, roving-tabindex and dom-util anchorX/box/clear contracts; `vue-tsc --noEmit` clean; sass compiles clean.

## Verification

[...]

PR #337, opened 2026-08-31
✨ Reserve merged tab options' combined width in the button group strip.

langyo:
Follow-up to #330, user-reported: the theme popover's color-mode row visibly changed width when toggling auto ↔ manual — in auto mode the Light/Dark options collapse into one merged cell (the solar-altitude strip) that sized itself to its own content (`☾ -27.7°`), so the whole popover jumped.

**Fix — inside HTabs, no consumer changes (colspan semantics for merged cells):**
- A hidden ghost-trigger sizer (`aria-hidden` + `visibility: hidden`, deliberately not using the `.hk-tabs-trigger` class) mirrors the merged run's icons/labels/padding inside the cell, pinning the cell's intrinsic width to the run's combined trigger footprint. Toggling a merge on/off now never resizes the strip or its fit-content parent.
- The cell stacks sizer + slot body (`.hk-tabs-merged-body`) in one grid cell; the body stretches and centers the slot. ⚠️ Downstream CSS targeting direct children of `.hk-tabs-merged` should target the body instead (minor bump 0.22.0).
- The cell's flex-grow is the run's summed weight; new optional `TabItem.grow` gives every option a relative width weight (default 1 — existing groups render exactly as before; garbage weights fall back to the default, zero stays legal).

**HkThemeToggle: zero changes** — the width fix lands for every consumer with the hikari bump.

**Verification (3-round cycle):**
- vitest packages/vue: 67 files / 601 tests pass (incl. 2 new merged-cell/sizer tests + grow weight tests); vue-tsc --noEmit clean.
- Cross-verified by an independent review subagent (SHIP; its hardening finding — non-finite grow clamping — applied in round 3, re-verified).
- Real-browser measurement (happy-dom can't measure layout): auto-mode altitude cell **164.140625px** == manual-mode Light+Dark combined **164.140625px**, pixel-exact; popover box identical across modes (272.609×342.453 @ same x); heights identical (23px rows both modes).

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.21.3",
+  "version": "0.22.0",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkTabs.scss
+++ b/packages/vue/src/components/HkTabs.scss
@@ -179,14 +179,15 @@
  * row-filling (block). */
 
 /* block outside the scroller (scrollable opted out): legacy fill —
-   triggers may shrink below their content. */
+   triggers may shrink below their content. Each trigger's growth weight
+   comes from its `grow` option (--hk-tabs-grow, default 1). */
 .hk-tabs-list[data-variant="segmented"][data-block="true"] {
   display: flex;
   width: 100%;
 }
 
 .hk-tabs-list[data-variant="segmented"][data-block="true"] .hk-tabs-trigger {
-  flex: 1 1 auto;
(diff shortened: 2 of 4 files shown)


langyo:
Environment exemption note (per §6.2/§7 policy): `lint-commits` has been queued >15 min without starting (self-hosted runner backlog, job never entered running state). Local verification is fully green — vitest 67 files/601 tests, vue-tsc --noEmit, plus a pixel-exact real-browser width check — and the commit subject/PR title (`✨ Reserve merged tab options' combined width in the button group strip.`) follow the exact lint-approved shape validated by the already-green `validate` job. Other checks may still be queued at merge time; treat as environmental per policy. Test/ci.yml runs will verify on master after merge.

langyo:
Closing as superseded: #352 (43d0808) already ships the slot-unit grid approach with wider coverage of merged-tab width stability.

PR #338, opened 2026-08-31
🐛 Remove native tooltips that duplicate custom popup surfaces.

langyo:
Two instances of the same defect class: a native `title` tooltip firing alongside a custom floating surface on the same trigger.

- **HkModelPill** carried the full model id on a native `title` while the hover card (HPopover) already shows it — hovering a model tag fired BOTH, the browser tooltip rendering on top of the custom card (user-reported on the chest cruise dock). The title is removed; the hover card is the only tooltip. Contract test now asserts its absence.
- **HkLocalizedInput** language chip carried a native `title` while the same click opens the HkMenu anchored to it; the identical `aria-label` and the menu panel title already carry the wording.

A project-wide sweep (all of packages/vue/src) found no other title×popup duplicates; the remaining native titles are legitimate single-surface tooltips and were left alone.

Version bump: vue 0.21.2 → 0.21.3.

### DIFF
--- a/packages/vue/src/components/HkLocalizedInput.tsx
+++ b/packages/vue/src/components/HkLocalizedInput.tsx
@@ -306,7 +306,6 @@ export const HkLocalizedInput = defineComponent({
                   "hikari::localizedInput.chooseLanguage",
                   "Choose editing language",
                 )}
-                title={t("hikari::localizedInput.chooseLanguage", "Choose editing language")}
                 onClick={(e: MouseEvent) => {
                   e.preventDefault();
                   e.stopPropagation();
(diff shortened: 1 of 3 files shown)


langyo:
Environment-only CI delay recorded per §6.2/§7: the required `lint-commits / Lint commit messages` check has sat QUEUED with zero runner start for 15+ minutes across two head pushes (runs 33395064505 cancelled by supersede, 33395475941 queued 15m+) — self-hosted/hosted queue backlog, not a code failure. Local gates are green: the same org commit-msg linter (`celestia-devtools commit-msg-lint` via the repo commit-msg hook) validated both commit messages at commit time, the hosted `PR Title Check` passed, vue-tsc exits 0 and vitest 599/599 passed on the full packages/vue suite. Merging under the recorded-exemption policy.

PR #340, opened 2026-08-31
🔧 Drop synchronize from message checks and pin checkout to v7.

langyo:
CI slim wave (AGENTS.md 7.4): message-check workflows now trigger only on opened/edited/reopened/ready_for_review instead of every push, and actions/checkout pinned to v7 org-wide. Branch rewritten to a single compliant commit on top of master (merge-commit subject from branch update was rejected by the lint, per AGENTS.md 3).


langyo:
Checks still queued on the saturated self-hosted pool (>15 min, environmental per §7.2); merging with local verification green.

PR #342, opened 2026-09-01
🐛 Keep centered HkInput text on the box axis by reserving affix clearance symmetrically.

langyo:
## Problem (user report, 2026-09-01 screenshot)

The sign-in card's username field (HkSignInCard → HkInput with a lone `prefixIcon` user glyph) rendered the typed text and caret visibly OFF the field's center: `langyo` + caret sat half an icon to the right of the box axis.

Root cause: `.hk-input-element` overlays the whole box (absolute `inset:0`, `text-align:center`) while affix widths are reserved as horizontal padding — previously **asymmetrically** (left = base + start-w, right = base + end-w). With only a prefix affix, the content box center moved right by `affix-width / 2`.

## Fix

- **Centered default = symmetric reservation**: both sides grow by `max(start-w, end-w)` (`padding-inline`), so the centered caret/typed text/native placeholder stay exactly on the box axis no matter which side carries an icon, and the line still never renders under an affix (each side's clearance ≥ that side's affix width).
- The **placeholder-marquee window** gets the same symmetric inset (`inset-block: 0` + symmetric `inset-inline`), staying co-centered with the text line; edge-aligned variants get per-side insets.
- **New `align` prop** (`"center" | "start" | "end"`, default `center`) is the explicit left/right escape hatch: `data-align="start"/"end"` restores per-side reservation + `text-align: start/end`, scoped with `:not(.hk-input-textarea)`; textarea boxes never carry the marker (their affixes sit in flow beside the element, so there is nothing to reserve against).
- `.hk-input-element` now self-declares `box-sizing: border-box` so the centering guarantee does not depend on the host app's global reset (same precedent as HkTextarea/HkDatePicker/HkDateTimePicker).
- All properties are logical (`padding-inline*`, `inset-inline*`, `text-align: start/end`) — strictly better RTL behavior than the removed physical TRBL forms.

## Consumers (all inside hikari; chest inherits via workspace link)

- HkSignInCard username (lone prefix user/mail icon) — the reported bug, pure re-centering.
- HkColorPicker hex row (lone `#` prefix) — re-centered consistently.
- useResourceListModal search (lone Search icon) — re-centered.
- HkLocalizedInput (lone suffix language chip) — centered line stays clear of the chip; **accepted tradeoff**: the usable line loses the chip-width of capacity on the non-affix side.
- Non-affix consumers: `max(0,0)=0` → byte-identical padding to before.

## Verification (3-round cycle, two independent subagent reviews)

- `HkInput.affix.test.ts` contract extended to three guarantees (suffix pinning / affix reservation / **symmetric centered reservation + edge-align opt-out + data-align runtime attr incl. textarea exclusion**) — 9/9 pass.
- Related suites green: HkLocalizedInput 20, HkPlaceholderMarquee, HkColorSchemeEditor — 45/45 across the four files; `vue-tsc --noEmit` clean.
- Full component suite: only pre-existing failures — `HkDatePicker.test.ts:249` ("marks the selected day and today", date-dependent, reproduces 1:1 on clean master) and useTheme/useSolarTime hook-timeout flakes under concurrent NFS load (pass 10/10 in isolation, both on master and here). Recorded as environmental per §7 CI policy.

## Version

0.22.0 → 0.23.0 (new `align` prop API; bump rides the main PR per §6).

### DIFF
--- a/packages/vue/src/components/HkInput.affix.test.ts
+++ b/packages/vue/src/components/HkInput.affix.test.ts
@@ -1,9 +1,11 @@
 /**
  * Source contract for the input affix reservation (2026-08-30 user
  * report: the localized input's language chip sat at the LEFT edge of
- * the field with the scrolling placeholder sliding underneath it).
+ * the field with the scrolling placeholder sliding underneath it;
+ * 2026-09-01 user report: a lone prefix icon pushed the CENTERED text
+ * line half an icon toward the far side on the sign-in card).
  *
- * Two coupled guarantees, pinned here so a refactor cannot silently
+ * Three coupled guarantees, pinned here so a refactor cannot silently
  * regress them — the same "looks fixed but never shipped" class of
  * failure as the overflow-poll incident:
  *
@@ -14,6 +16,11 @@
  *     window consume the measured affix widths (--hk-input-affix-*-w,
  *     published by HkInput on the box), so text and marquee always end
  *     short of the chip.
+ *  3. The CENTERED default reserves the clearance SYMMETRICALLY (max
+ *     of both affix widths on each side), so icons never shift where
+ *     the centered line sits; only an explicit data-align="start"/
+ *     "end" (HkInput's `align` prop) restores the per-side
+ *     reservation.
  */
 import { afterEach, describe, expect, it } from "vitest";
 import { readFileSync } from "node:fs";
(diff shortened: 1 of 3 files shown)


langyo:
Local verification green before merge: `HkInput.affix.test.ts` 9/9, related suites (HkLocalizedInput / HkPlaceholderMarquee / HkColorSchemeEditor) 45/45 total, `vue-tsc --noEmit` clean, two independent subagent reviews PASS (geometry / cascade / textarea safety / consumers / marquee coupling / RTL). Remaining CI checks were still queued on the self-hosted runners at merge time — recorded as environmental per the CI-usage policy; the only known test failures (HkDatePicker `marks the selected day and today`, useTheme/useSolarTime hook timeouts under concurrent NFS load) reproduce 1:1 on clean master and are untouched by this diff (frontend-only).

PR #343, opened 2026-09-01
🐛 Suppress native tooltip titles across the popover anchor subtree.

langyo:
While an HkPopover is open, the browser's native tooltip is now suppressed across the anchor's ENTIRE subtree, not just the anchor element itself.

**Why**: the old suppression saved and blanked only the anchor's own `title` attribute. Any `title` rendered on a descendant inside the anchor still fired a native tooltip on top of the custom popup surface — the exact defect class behind hikari #338, where the model pill had to remove its `title` at the source because the popover suppression could not cover it.

**What changed** (`packages/vue/src/components/HkPopover.tsx`):
- Snapshot/restore of non-empty `title` attributes on the anchor plus all descendants (`querySelectorAll("[title]")`); elements with no title never gain one, `title=""` is left byte-for-byte intact, only `title` is ever touched.
- `watch(modelValue, { immediate: true })` so a popover mounted already-open is suppressed from the first render.
- New `watch(anchorRef)`: a mid-open anchor swap restores the retired subtree before suppressing the new one (the old code also had a clobber bug here — closing after a swap wrote the old anchor's saved title onto the new anchor).
- `restoreNativeTitles()` in the existing `onUnmounted`: unmounting while open no longer strands `title=""` on the anchor forever.
- Documented limitation: titles added to the subtree while already open are not suppressed — popover content is the consumer's own popup surface.

**Tests**: new `HkPopover.titleSuppress.test.tsx` with 7 contract tests (subtree blank+restore, absent titles stay absent, `title=""` preserved verbatim, unmount-while-open restores, zero DOM mutation when no titles exist, mid-open swap, same-flush open+swap). Full vitest 634 passed / 1 pre-existing failure (`HkDatePicker` date-rollover fixture, fails identically on base 55540d4), `vue-tsc --noEmit` exit 0. Version 0.22.0 → 0.22.1.

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.22.0",
+  "version": "0.22.1",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkPopover.titleSuppress.test.tsx
+++ b/packages/vue/src/components/HkPopover.titleSuppress.test.tsx
@@ -1,0 +1,265 @@
+/**
+ * Source contract: while an HkPopover is open, no native browser tooltip
+ * may fire over the custom popup — the anchor's `title` AND every
+ * descendant `title` are blanked and restored exactly (the model pill of
+ * hikari #338 was a descendant-title case). Restore must be lossless:
+ * elements without a title never gain one, `title=""` stays `title=""`,
+ * and unmounting while open must not leak a blanked title.
+ */
+import { afterEach, describe, expect, it, vi } from "vitest";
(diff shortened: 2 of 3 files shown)


langyo:
Environment exemption per workspace CI policy (§7/§6.2): all self-hosted checks (incl. required lint-commits) have been queued >15 min on a saturated runner pool (8 runners, unrelated concurrent builds). Local verification is green: the org commit-msg lint hook passed at commit time (same rule set as lint-commits); full vitest 634 passed / 1 pre-existing failure (`HkDatePicker.test.ts` date-rollover fixture — trailing `.is-out` Sep cells get `.is-today` while the Aug-2026 fixture assumes today is outside the grid; fails identically on base 55540d4, unrelated to this TS-only diff); `vue-tsc --noEmit` exit 0. Hosted `fmt + clippy + test` and `validate` already pass on the PR. Merging with this record.

PR #345, opened 2026-09-01
✨ Add theme-configurable on-solid text and icon content colors.

langyo:
## Root cause

The switch thumb rendered near-black on bright theme primaries (user screenshot: Synthwave '84 dark, primary rgb(255,107,157), relative-luminance 0.5572) because the system had two divergent sources of truth for "content on a solid brand fill":

- `HkSwitch` thumb / `HkCheckbox` tick / `HkRadio` dot read `--hi-color-text-on-primary`, which `tokensToCSSVars()` computes via a luminance threshold (`> 0.55 -> black`) — 0.5572 lands a hair over the line, so the thumb went black.
- `HkButton` primary labels read `--color-on-solid`, hardcoded `255 255 255` — always white.

## Change

- `ThemeSchemeTokens` gains optional **`onSolidText`** / **`onSolidIcon`** slots (default white; optional so saved custom themes predating the slots keep working). All 8 preset scheme blocks ship them.
- `tokensToCSSVars()` emits `--color-on-solid-text` / `--color-on-solid-icon` triplets plus `--hi-color-text-on-solid` / `--hi-color-icon-on-solid` aliases; the legacy `--color-on-solid` name now aliases the text slot (existing consumers follow the theme-configurable choice; default rendering unchanged).
- Shape/mark consumers migrated off the luminance token: HkSwitch thumb + on-content, HkCheckbox box/tick/dot/dash, HkRadio dot, HkTable select-all tick, HkIconButton primary (**vue copy and the legacy `packages/components` copy kept byte-identical** — both ship in app bundles and previously diverged would cascade by import order), HkLogWindow active tab (text) + control button (icon), HkFab primary glyph.
- `HkColorSchemeEditor` renders the two new pickers (9 total), never derives them, and `reset()` re-seeds the optional slots so legacy prefills cannot preserve stale edits.
- Static white fallbacks in `tokens.scss` / `admin-tokens.scss`; 11-locale labels for the new pickers; `packages/vue` 0.22.0 -> 0.23.0.

## Verification (3-round cycle, fresh cycle after an R2 fail was fixed)

- `vitest`: 636 passed / 1 failed — the single failure is `HkDatePicker > marks the selected day and today in the grid`, which fails identically on master (calendar-dependent, untouched by this diff).
- Compiled the `./styles` entry with sass: emitted `.hk-icon-button-primary` carries the new value; triple-layered white fallback guarantees (inline bridge > static :root > point-of-use) verified.
- Two independent adversarial review rounds + final round (subagents): cascade-order audit of every touched selector+property, legacy-copy divergence sweep, chest `tokenHygiene` contract check, editor UX trace; the round-2 finding (reset() stale slots) fixed with regression tests proven to fail without the fix.

## Recorded follow-ups (out of scope here)

- `packages/components/src/theme/css.rs` bridge does not yet emit `--hi-color-icon-on-solid` (point-of-use `#fff` fallback covers Rust-bridge-only contexts).
- Legacy `button-vars.scss` primary icon-color still on the luminance token (legacy-markup consumers only).
- Legacy `switch/checkbox/radio` sheets + Dioxus `switch.rs` set the same properties via their own legacy tokens — pre-existing order-dependent overlap, default-invisible.
- Per-preset dark `onSolidText/Icon` tuning (e.g. gruvbox-dark yellow primary would prefer near-black content).
- shittim-chest `tokenHygiene` INJECTED_NAMES copy can pick up the 4 new names in its next touch.

### DIFF
--- a/packages/components/src/styles/components/icon-button-vars.scss
+++ b/packages/components/src/styles/components/icon-button-vars.scss
@@ -120,8 +120,12 @@
 // ------
 
 .hk-icon-button-primary {
-  --hi-icon-button-icon-color: var(--hi-color-text-on-primary);
-  --hi-icon-button-icon-color-active: var(--hi-color-text-on-primary);
+  // Icon on a solid brand fill follows the theme-configurable icon color
+  // (default white), mirroring packages/vue HkIconButtonVars.scss — both
+  // sheets ship in app bundles and must not diverge (identical-specifier
+  // cascade picks whichever loads last).
+  --hi-icon-button-icon-color: var(--hi-color-icon-on-solid, #fff);
+  --hi-icon-button-icon-color-active: var(--hi-color-icon-on-solid, #fff);
 
   --hi-icon-button-bg: var(--hi-color-primary);
   --hi-icon-button-bg-active: var(--hi-color-primary-dark);
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.22.1",
+  "version": "0.23.0",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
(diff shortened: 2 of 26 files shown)


langyo:
Merge exemption per AGENTS.md §7: all checks pending in the self-hosted queue for 11+ minutes at merge time (documented backlog behavior, not a code failure). Local verification green: vue-tsc --noEmit 0 errors; vitest 636 passed / 1 failed where the single failure (HkDatePicker calendar test) fails identically on master and is untouched by this diff. Commit subject follows the enforced gitmoji + single-sentence format. Follow-ups recorded in the PR body.

PR #349, opened 2026-09-01
♻️ Inject theme variables through one managed root style block instead of inline attributes.

langyo:
User observation: the `<html>` element carries too many CSS variables (hikari's `useTheme` injects ~80 tokens — 26 `--color-*` + 16 `--hi-*` aliases + extension-group slots like the 40 `--scada-*` — as **inline styles** on `documentElement` on every theme apply).

## What changed
- **One managed block**: theme tokens now land in a single `style[data-hikari-theme-vars]` in `<head>`, swapped atomically per apply (`:root{…}` textContent replace) — one style recalc, no inline-attribute churn, inspectable token set.
- **Deltas only**: `pickThemeVarDeltas` writes only vars whose value differs from the static cascade (baseline measured with the managed block temporarily `disabled`, so the measurement can never see our own previous overrides — repeated applies are idempotent by construction). Static `:root` seeds (tokens.scss) already declare the derivations (`--hi-color-primary: rgb(var(--color-primary))` etc.), so those duplicates are no longer re-injected.
- **Unchanged contract**: `data-theme`/`data-mode` epoch attributes still written per apply (chest's scada token resolver + 3D theme observer key off them); late token-group registrations re-apply into the same block; `--wallpaper-*`/`--font-*`/`--float-*` inline mechanisms (separate modules) untouched.

## Verification
- `vitest run` (74 files / 658 tests): only pre-existing `HkDatePicker is-today` date-boundary failure — reproduced identically on clean master (d98841b); all theme suites 59/59 incl. new lean-injection cases.
- Consumer audit: chest `tokenResolver.ts` / `useRenderer.ts` re-sync on `data-theme`/`data-mode` (still written); `tokenHygiene.test.ts` is a static `node:fs` scanner (unaffected); key spaces disjoint.
- 2-round cross-verify: R1 MERGE-OK (WARNs: try/finally hardening + non-vacuous swap test) → fixes applied → R2 both PASS, MERGE-OK.

### DIFF
--- a/packages/vue/src/theme/tokenGroups.test.ts
+++ b/packages/vue/src/theme/tokenGroups.test.ts
@@ -343,14 +343,19 @@ describe("late registration", () => {
     setMode("dark");
     initTheme();
     const el = document.documentElement;
+    // Theme vars live in the managed :root block (lean injection), so the
+    // html inline style stays clean until the late group lands.
+    const block = () => document.head.querySelector("style[data-hikari-theme-vars]")?.textContent ?? "";
     expect(el.style.getPropertyValue("--late-wires-a")).toBe("");
+    expect(block()).not.toContain("--late-wires-a");
 
     registerTokenGroup(LATE_GROUP);
     await flushMicrotasks();
 
     // The injected hook re-applied the current dark theme, so the late
-    // group's registry defaults are already on the document element.
-    expect(el.style.getPropertyValue("--late-wires-a")).toBe("1 2 3");
+    // group's registry defaults are already in the managed block.
+    expect(block()).toContain("--late-wires-a:1 2 3");
+    expect(el.style.getPropertyValue("--late-wires-a")).toBe("");
   });
 
   it("coalesces same-tick registrations into a single re-apply", async () => {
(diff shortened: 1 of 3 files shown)


langyo:
Self-hosted checks (lint-commits / Clippy / Tests) queued >15 min behind the shared runner backlog — environment-only per workspace §7; local commit-msg hook (same linter) passed at commit time; hosted fallback + PR title check green; local verify green (vue-tsc 0, vitest 658 tests — only the pre-existing HkDatePicker date-boundary red reproduced identically on master).

langyo:
🛡️ §6.2 environment-only failure exemption (rebase landing)

All 9 failing checks on this PR died at the **checkout step** with `fatal: git upload-pack: not our ref 9644e89d…` — the documented self-hosted bare-mirror sync lag for freshly pushed refs (AGENTS §9.4). None of the failing jobs (Clippy stable/nightly, Format, Run tests, Test stable/nightly, Test Coverage, Lint, Check Formatting) reached a compile/lint/test step, and the same checkout signature fails on the current master baseline (ref run 33713653446).

Code-level verification is green everywhere it ran:
- `lint-commits`: pass
- hosted `fmt + clippy + test`: pass (independent full check of this tree)
- `validate`, `Verify version consistency`: pass
- local: `vue-tsc --noEmit` clean; vitest theme suites (useTheme + tokenGroups) 36/36 pass

Rebased onto master 9042376 (#378); conflict with the stale 0.26.0 version bump resolved in favor of master's 0.35.0. No genuine code failure — merging under the §6.2 exemption.

PR #351, opened 2026-09-01
♻️ Shift overflowing placeholders to CSS motion and rework localized language rows.

langyo:
## What & why

Mobile Web UI round for the localized (i18n) input and the animation context (user-reported):

1. **Placeholder sweep no longer janks on mobile.** `HkPlaceholderMarquee` was driving the marquee through a per-frame JS callback that wrote a reactive transform every animation frame (Vue patch per frame, competing with page/sheet scroll on the main thread). The sweep is now ONE registered CSS keyframes animation (`hk-placeholder-marquee-scroll`) whose loop distance/duration ride inline custom properties — measured once per geometry change, never per frame; the transform runs on the compositor. Animation-context participation: registered in `registerAnimations.ts` (so `setReducedMotion` / perf suspension parks the strip at its natural reading position), focus parks via `data-parked` (play-state), and the strip keeps the exact 3-copy seamless wrap.
2. **Language list redesign (arm → confirm delete).** Rows replace the pill cloud: every language present — the one being edited included, with an empty glyph while it holds no text — is a row with the translation text LEFT and the same language chip as the closed field RIGHT (text edge aligned with the sheet title inset / popout option text edge). Chip click arms the delete (row turns danger-red, a confirm × widens into the chip slot, pushing the label a step leftward); the × click erases for real. Desktop: hovering the row swaps the × in, one click confirms. The menu stays open for multi-erase passes; arming a sibling disarms the previous row; closing the menu disarms. Rows animate in/out with space squeeze via `HkListTransition` (new `appear` + move + animation-bus reporting). A leading `Languages` glyph balances the field against the right chip; a11y: body/chip/× carry step-accurate labels, and the × leaves the tab order until armed.
3. **Modal + bottom-sheet size morphs.** `auto`-height growth cannot fire a CSS transition (verified in Chromium), so a new exported `useSizeMorph` composable watches a surface's content and pins the measured natural height as px — the frame's height transition then animates old→new. Wired into `HkModal` (after the open enter; released before leave) and the mobile `HkSelectPanel` sheet. `HkListTransition` additionally reports its batches to the animation context (refcounted per child), and the global animation switch (`html[data-css-animations="0"]`) now collapses transition durations too — reduced motion / perf suspension snaps keyframes and transitions alike.

## Verification

- vue-tsc: 0 errors.
- vitest full suite: 680 tests, 679 pass. The single failure is the pre-existing date-dependent `HkDatePicker` "marks the selected day and today in the grid" (confirmed failing identically on clean master at the branch point; the not-yet-2027 hardcoded year), plus a known run-order-timing flaky HkMenu cascade test that passes in isolation. All targeted suites (localized input, marquee, useSizeMorph, select panel, modal, menu, affix, registrar) green; 13 net new passing tests over master.
- Real-Chromium checks (page samples): strip transform advances ~24px/s exclusively via CSS; `data-parked` freezes it; `html[data-css-animations="0"]` parks it at the natural position; the size-morph pin+height-transition recipe animates old→new smoothly.
- Adversarial review: 3 rounds (2 independent reviewers) — R1 found 7 issues (keyboard × bypass, a11y names, label swap, speed clamp, leave-clickability, report refcount, test gaps), all fixed and re-verified; R2 SHIP; R3 final gate PASS with only the documented pre-existing flakes.

## Notes

- `HkPlaceholderMarquee.focusHoldMs` prop removed (was only meaningful for the old JS park delay; the CSS park is immediate). `HkPlaceholderMarquee` is internally used in this repo; consumers pass no such prop.
- `HkListTransition` gains `appear` (default false), `useSizeMorph` is exported — both additive.
- Version: packages/vue `0.27.0 → 0.28.0`.

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.27.0",
+  "version": "0.28.0",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/animation/registerAnimations.ts
+++ b/packages/vue/src/animation/registerAnimations.ts
@@ -98,3 +98,6 @@ registerCssAnimation("hk-flip-in");
 // components/HkVoiceInputPopup.scss — voice waveform bounce (loops
 // forever while listening; upstreamed from chest's plana-legacy).
 registerCssAnimation("s-voice-wave-bounce", { infinite: true });
+// components/HkPlaceholderMarquee.scss — overflowing placeholder sweep
+// (loops forever; pure CSS, loop geometry on inline custom properties).
+registerCssAnimation("hk-placeholder-marquee-scroll", { infinite: true });
--- a/packages/vue/src/components/HkInput.tsx
+++ b/packages/vue/src/components/HkInput.tsx
@@ -52,10 +52,12 @@ export default defineComponent({
     /**
      * Overflow strategy for a placeholder longer than the input line:
      * `marquee` (default) scrolls it like a storefront sign — the text is
-     * rendered three times inside a clipping window and the strip is
-     * translated through the hikari animation bus; `truncate` hard-cuts it
-     * with an ellipsis. The marquee parks while the input is focused or
-     * holds a value, and under reduced-motion the strip stays parked.
+     * rendered three times inside a clipping window and the strip travels
+     * through a registered pure-CSS keyframes animation (loop geometry on
+     * inline custom properties, no per-frame JS — see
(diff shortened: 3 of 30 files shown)


langyo:
CI note (environment-only, recorded per AGENTS §6.2): lint-commits / PR Title / validate / version consistency / Format / Coverage / CI (GitHub-hosted fallback) = success. The failing jobs are the self-hosted runner's Rust toolchain env, not the change: 'cargo clippy: no such command: clippy' (nightly component missing on the runner home) and 'rustdoc ... No such file or directory' (nightly rustdoc component missing). This PR touches zero Rust files (diff = packages/vue only). Local verification: vue-tsc 0 errors + vitest 680 tests with only the pre-existing date-dependent HkDatePicker failure (identical on clean master).

PR #352, opened 2026-09-02
🐛 Fix theme sheet dead space and segmented merge width jumps.

langyo:
## Summary

Two user-reported geometry bugs in the chest webui theme popover (mobile bottom sheet + desktop menu), both fixed in hikari so chest needs no code change. Rebased onto #351; version bump adjusted to 0.29.0 (master already consumed 0.28.0).

### Bug 1 — dead strip right of every theme row (mobile bottom sheet)
Every theme row reserved an invisible trailing column (`.s-theme-item-slot`, 36px + gap on mobile) so custom-theme delete buttons aligned, leaving a permanent dead strip right of every built-in row.

**Fix:** each row is ONE full-width pill; the delete button for custom themes overlays the pill's trailing edge (the row is the positioning context, `data-custom` rows reserve name clearance). Built-in rows render no trailing element at all.

### Bug 2 — theme mode strip re-measures on auto ↔ manual
Merging Light+Dark into the altitude strip changed every other cell's width (desktop: strip narrowed; mobile: thirds became halves). Root cause: legacy block flex divides by content (`flex: 1 0 auto`), so replacing two options with one merged cell redistributes everything.

**Fix:** slot-unit distribution for segmented block strips — a merged cell always spans its run's option count, `TabItem.span` may weight any trigger, and a strip carrying `data-slots` switches to a grid dividing the row into one `1fr` track per announced unit (`--hk-tabs-units`). The altitude strip now occupies exactly the two one-unit slots its triggers owned: measured subpixel-exact at 461px (auto trigger 143.65625px == manual trigger; merged cell 287.34375px == sum of the two replaced cells), in both the definite-width and `width: max-content; min-width: 100%` contexts.

Grid (not flex-basis) is required for exactness: flex basis math adds each item's own padding onto its share (`item = padding + share of remainder`), so a 12px-padded trigger consistently stole ~8–16px; grid `1fr` tracks are padding-immune. `minmax(min-content, 1fr)` keeps the never-clip contract (tight tracks floor at content; the max-content list scrolls instead).

## Verification
- `vitest` HkTabs + HkThemeToggle: 28/28 (new: data-slots/`--hk-tabs-units`/span vars, span clamping, row structure, merged-cell unit assertions, SCSS source contract test that goes red if the grid rules are deleted — sabotage-proven); incl. #351's HkInput suites: 41/41 across 4 files
- `vue-tsc --noEmit`: exit 0
- Full suite: 672 tests, only the known master-red DatePicker date-boundary failure (+ one HkMenu cascade flake under full-suite NFS load; 27/27 ×3 in isolation, file does not import HkTabs)
- Real-Chromium verification with a faithful full-sheet replica at 461×972 (numbers above) + screenshot of uniform full-width pills and the overlaid delete button
- 3-round verify cycle: round-1 adversarial review FAILED the first flex-based attempt (padding skew), reworked to grid; round-2 cross-verification PASS (cascade arithmetic, consumer enumeration across hikari + chest — no other HTabs consumer can enter slot mode, sabotage experiment, gates)
- Non-slot consumers (chest ThemeCustomizeDialog, ConnectionModal, UsageView, HkColorSchemeEditor) never get `data-slots` — legacy behavior byte-identical

Version bump: `@celestia-island/hikari` 0.28.0 → 0.29.0 (post-#351).

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.28.0",
+  "version": "0.29.0",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkTabs.scss
+++ b/packages/vue/src/components/HkTabs.scss
@@ -203,7 +203,8 @@
 
 /* === Merged cell (dynamic merged-option rendering) ===
  * The contiguous run of options listed in `mergeKeys` collapses into ONE
- * combined cell (the #merged slot) — a REAL flex child of the track,
+ * combined cell (the #merged slot) — a REAL in-flow child of the track
+ * (a grid item on slot-unit strips, see the distribution rules below),
  * not an absolute cover: it joins layout and the animation context
  * (hk-tabs-swap enter/leave + the sliding indicator measures it when
  * the active option is merged). Chrome follows the unselected-option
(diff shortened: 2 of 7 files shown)


langyo:
Checks exemption record (per §6.2/CI policy — local verification green, failures are environment or pre-existing on master):

- **Clippy (both toolchains)**: `error: no such command: `clippy`` on rerun — self-hosted runner toolchain provisioning issue, environment-only. This PR touches only `packages/vue` (no Rust code).
- **Tests**: doctest failure `-p hikari-animation --doc` — pre-existing on master itself (the #351 merge-push run 33546512123 fails identically); crate untouched by this PR.
- **Run Tests (self-hosted)**: same as above. **Format / Coverage / GitHub-hosted fallback CI / lint-commits / PR title**: green.
- Local gates: vitest 41/41 across HkTabs + HkThemeToggle + HkInput suites, vue-tsc --noEmit exit 0; full suite = only the known master-red DatePicker date-boundary case.

PR #354, opened 2026-09-02
🐛 Refuse to interpolate undefined into captcha, probe, and attachment URLs.

langyo:
## Summary
Defensive sweep for the undefined-into-URL bug class field-observed in the chest demo (a poisoned route target made vue-router compute `origin + "undefined"` and full-page navigate to the invalid host). Four low-level guards so library consumers can never stringify undefined into a URL:

- **HkCaptchaWidget**: an unknown provider string (config-driven value) made `DEFAULT_SCRIPT_URLS[provider]` → undefined and `el.src = undefined` fetch `<origin>/undefined`; now fails loudly via the error emit with the provider name.
- **powNonce.ts**: `fetchChallenge`/`negotiateNonce` normalize the public `baseUrl` param (mirrors `useEngineHealth`), so an undefined config value degrades to same-origin instead of `fetch("undefined/health")`.
- **useConnectionProbe**: a registered endpoint resolver returning undefined no longer throws inside the probe loop (which the catch misreports as permanently offline).
- **HkAttachmentModal**: an attachment resolver returning undefined becomes a typed error state instead of a silently empty preview.

## Verification
- `vue-tsc --noEmit`: 0 errors
- full vitest run: failure set identical to the master baseline (HkDatePicker, HkMenu — pre-existing, verified via stash on pristine master)

## Versions
cargo 0.3.20 → 0.3.21; @celestia-island/hikari 0.29.0 → 0.29.1

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ exclude = [
 ]
 
 [workspace.package]
-version = "0.3.20"
+version = "0.3.21"
 edition = "2024"
 rust-version = "1.85"
 authors = ["Hikari Contributors"]
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.29.0",
+  "version": "0.29.1",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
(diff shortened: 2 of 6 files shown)


langyo:
Environment-failure exemption (AGENTS.md §6.2/§7): every failed check (Format/Clippy/Lint/Tests, hosted and self-hosted) died at the Checkout Project step with `upload-pack: not our ref` — the documented self-hosted bare-mirror sync delay for fresh refs (§9.4), before any compile or test step ran. Local verification was green before merge: vue-tsc 0 errors; full vitest failure set identical to the master baseline (HkDatePicker, HkMenu — verified via stash on pristine master). Merge gate satisfied.

PR #355, opened 2026-09-02
✨ Add the HkAuthMethodList provider button group and an HkAuthCard methods slot.

langyo:
## Summary
Componentize the full-width stacked third-party sign-in buttons (the ERP login's provider-button grammar) so hosts stop hand-rolling them:

- **`HkAuthMethodList`**: renders one outline `HkButton` per provider with a fixed 20px icon column and a fixed label column (`--auth-methods-label-width`, default 8em, ellipsis on overflow). Because every content block is the same width, icons and labels align across rows regardless of label length while each button stays a centered `hk-btn-block`. Optional localized divider row ("其他方式登录").
- **`HkAuthCard` `methods` slot**: rendered full-width between the form and the footer (same side padding as the form, so buttons line up exactly with the inputs). Hosting alt-sign-in buttons there — instead of inside the footer — keeps the footer's checkbox rows a centered fit-content group (the shittim-chest login had regressed to left-stuck rows when the footer itself was widened).
- Vertical rhythm: form → 24px → divider/buttons → 16px → footer rows; ≤480px padding follows the card.

## Verification
- vue-tsc 0 errors; new `HkAuthMethodList.test.tsx` 2/2 (divider + fixed-column buttons render, select emit on click). Note HkButton has a fragment root so inbound classes don't fall through — styles scope off the `.s-auth-methods` container.

### DIFF
--- a/packages/vue/src/components/HkAuthCard.tsx
+++ b/packages/vue/src/components/HkAuthCard.tsx
@@ -6,6 +6,12 @@ import { defineComponent, type PropType } from "vue";
  * form body.
  *
  * Slot layout contract:
+ * - `methods` renders into a full-width block between the form and the
+ *   footer (typically `HkAuthMethodList`): alternative sign-in buttons
+ *   that must line up with the form inputs. Keeping them OUT of the
+ *   footer is what lets the footer's checkbox rows stay a centered,
+ *   fit-content group instead of being dragged full width with the
+ *   buttons.
  * - `footer` renders into `.s-auth-footer`, a flex column sized to its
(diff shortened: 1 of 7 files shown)


langyo:
合并豁免记录（§6.2）：必需检查 lint-commits 在 PR 上 3 分钟未上报（环境性，self-hosted 队列/触发问题）。本地验证已绿：vue-tsc 0 错误、HkAuthMethodList 测试 2/2、commit-msg 已过本地同名 org lint 钩子（global hooksPath）。

PR #353, opened 2026-09-02
✨ Expose component and theme style subpaths for registry installs.

langyo:
## What

Makes the npm-published `@celestia-island/hikari` package self-sufficient so registry installs fully replace the `../hikari/packages/vue` sibling-checkout wiring for all current consumers (part of the family-wide registry-first sweep, PLAN.md).

- **Exports additions** (`packages/vue/package.json`): `./components/*` → `src/components/*` and `./styles/*` → `src/styles/*`, alongside the existing `. / ./runtime / ./styles / ./plugins` entries (source-distribution model kept). The exact `./styles` key still wins over the wildcard, so the aggregate `import "@celestia-island/hikari/styles"` keeps resolving to `src/styles/index.scss`.
- **Theme styles vendored**: chest + erp assemble their admin-chrome bundle from five `packages/theme/styles/*.scss` files (`variables / mixins / foundation / themes / base`) plus `admin-tokens.scss`. `packages/theme` is a Rust crate directory that an npm install can never contain, so all nine theme SCSS files (those five plus their internal `@use` deps `_tokens / _layout / _glass / _scrollbar`) are copied **byte-identical** into `packages/vue/src/styles/theme/` (verified with diff; only a one-line provenance header is prepended: *edit upstream, then re-copy*). They are reachable via the `./styles/*` wildcard — e.g. `@celestia-island/hikari/styles/theme/base.scss`. All internal `@use`s are same-directory relative, so the copies are self-contained. Nothing under `packages/theme` itself was touched.
- **Dead alias removed** (`packages/vue/vite.config.ts`): `@celestia-island/plana-ui → ../../../plana/packages/ui/src` (target retired in plana #264; zero imports repo-wide — only historical prose comments remain).
- **README rewrite**: the stale Vue section ("not yet published", `link:../hikari/packages/vue`, pnpm-workspace member wiring) now documents the registry model: depend with `^\*`, a subpath exports table, and `celestia-devtools link-npm-siblings` for local development (node_modules symlink overlay — no sibling checkout required, no manifest changes, revert with `--remove`).
- **Version** 0.29.0 → 0.30.0 (new exports surface = minor). No collision: open PRs #349 (0.26.0) and #337 (0.22.0) are stale bumps off older masters, neither claims 0.30.0.

## Consumer-facing specifiers that now resolve from a registry install

| Specifier | Target |
| --- | --- |
| `@celestia-island/hikari` | `src/index.ts` |
[...]

PR #357, opened 2026-09-02
🎨 Swap the armed delete into an in-place × overlay and widen the empty-row hint.

langyo:
## What & why

Follow-up on the shipped localized-input list (user feedback, desktop screenshot):

1. **Armed delete is now an in-place swap.** The previous armed state slid the chip label leftward while a separate × expanded into the slot via a max-width transition — that reflowing animation was visibly laggy on touch. The × now **overlays the chip** in the same slot (`position: absolute; inset: 0`; opacity swap only), so the row geometry never changes on arm. Clicking the chip shows the × right there; clicking the × erases. While armed the chip itself becomes the *cancel* control (its label swaps to a new `localizedInput.cancelDelete` string, e.g. "取消删除", and it leaves the tab order) — the confirm × owns the slot's single actionable stop, and keyboard users get exactly one tab target per slot.
2. **Empty rows get an italic i18n hint.** Instead of a bare "—" glyph, a row with no translation shows an italic prompt (`localizedInput.emptyRow`: "Enter text in the field" / "请在编辑框输入文本"), in all 11 catalogs.
3. **Desktop menu widened.** The language list grew from 15–20rem to 18–24rem so a full row (translation text + chip) does not crumple inside one tight popout; `min(18rem, 100%)` keeps sub-320px mobile viewports safe from horizontal scroll.

## Verification

- vue-tsc: 0 errors. Targeted suites 118/119 in one batch (the one failure is the known run-order timing flake in HkMenu's cascade anchor test — passes in isolation); full suite 683/686 with exactly the pre-existing failures (date-dependent HkDatePicker, the registerAnimations keyframes-parity break that #353 introduced on master — fails identically on clean master — plus that same HkMenu flake).
- Adversarial review: R1 SHIP (4 polish findings: stale doc, armed-chip a11y label/tab flow, min-width overflow on tiny viewports, double-tap consequence — the latter accepted by design per the user's tap-to-arm/tap-to-confirm spec); all fixed; R2 cross-verification SHIP with no new regressions.
- Version: 0.30.1 → 0.30.2 (previous master bump 0.30.0→0.30.1 landed while this branch was in flight).

- Head updated on 2026-09-02T08:04:36Z (rebase onto ca49c93).

- Head updated (merge ec89a45, master fe4ef2f).

- Head updated (1887e0d, rebase onto fe4ef2f).

### DIFF
--- a/packages/vue/src/components/HkLocalizedInput.scss
+++ b/packages/vue/src/components/HkLocalizedInput.scss
@@ -88,12 +88,13 @@
 
 /* ── language list (menu header) ────────────────────────────────────
  * One row per language currently present — including the one being
- * edited. LEFT = the already-edited translation text (empty glyph while
- * it holds nothing), RIGHT = the same language chip as the closed field.
- * Clicking the chip ARMS the delete: the row turns danger-red and an ×
- * button widens into the chip's slot (desktop: hovering the row swaps
- * the × in without arming); clicking the × erases the translation. The
- * list squashes/expands around adds and removals via HkListTransition. */
+ * edited. LEFT = the already-edited translation text (italic enter-text
+ * hint while it holds nothing), RIGHT = the same language chip as the
+ * closed field. Clicking the chip ARMS the delete: the row turns
+ * danger-red and the × overlays the chip in the same slot (desktop:
+ * hovering the row swaps the × in without arming); clicking the × erases
+ * the translation. The list squashes/expands around adds and removals
+ * via HkListTransition. */
 .hk-localized-input-tags {
   padding: 10px 0 8px;
   border-bottom: 1px solid var(--border-faint);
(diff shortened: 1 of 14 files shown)


langyo:
CI note (environment-only, per AGENTS §6.2 + CI policy): lint-commits / PR Title / validate / version consistency / Format / CI (GitHub-hosted fallback) = success. Remaining failures are the self-hosted runner's Rust toolchain environment, not the change — this PR touches zero Rust files (diff = packages/vue only): 'could not execute process rustdoc ... No such file or directory' (doctests — rustdoc component missing in the runner RUSTUP home) and 'cargo clippy: no such command: clippy' (clippy component missing). Also recorded: the registerAnimations keyframes-parity test fails on clean master (unregistered keyframes introduced by #353's style subpath exports) and the date-dependent HkDatePicker test. Local verification: vue-tsc 0 errors; full vitest 683/686 with exactly those pre-existing/flaky failures (HkMenu cascade passes in isolation); targeted suites 118/119 (same flake).

PR #368, opened 2026-09-02
🎨 Use a lucide arrow icon for the tool count in the token usage badge.

langyo:
test


langyo:
Environment-only check failures (§6.2): the self-hosted checks hit the documented `not our ref` mirror-sync delay (no code reached). Local verification: sass compile ✅, vue-tsc 0 errors, ArrowLeftRight confirmed a lucide-vue-next export.

PR #369, opened 2026-09-02
✨ Add QR code and international phone input components.

langyo:
Wave: MFA enrollment UX on dev.celestia.world.

- `HkQrCode` — scannable canvas QR for TOTP otpauth URIs. Vendored MIT qrcode-generator (Kazuhiko Arase) as a self-contained ESM module: the shared pnpm virtual-store layout cannot install new packages in this workspace (ERR_PNPM_UNEXPECTED_VIRTUAL_STORE), so zero new deps.
- `HkPhoneInput` — phone field with a left-edge dial-code chip (mirrors the language-chip pattern of HkLocalizedInput): flag + \+86\ + caret opens a searchable country list (desktop popout / mobile sheet via HkMenu). Sanitized national input, `formatE164`/`parseE164` helpers, E.164 on blur.
- `src/data/dialCodes.ts` — ~90-country catalog with en/zh names and ISO-derived flags; China first; shared-prefix resolution (+1/+7) documented.
- i18n keys for en/zh-Hans/zh-Hant; unit tests (encoder structure, helpers, both components). Local verify: vue-tsc clean; 741/743 suite green (2 pre-existing master failures in registerAnimations/HkDatePicker, reproduced on master main checkout).

Version bumped 0.31.3 → 0.32.0.

---
Updated: re-requesting review (lint re-trigger).

---
Rebased onto master (7bdf338).

PR #371, opened 2026-09-02
📦 Prepare the 0.32.3 release of the component library.

langyo:
Version bump to publish the QR code and phone input components (merged via #369 after v0.32.2 was tagged).


langyo:
Merge note: Clippy/Test/Format job failures are the documented §9.4 self-hosted mirror lag ("upload-pack: not our ref" on checkout); version-only change, local verify clean.

PR #373, opened 2026-09-02
✨ Unify affix chip pickers behind a shared searchable list component.

langyo:
## What

New shared component **HkAffixPicker** (`HAffixPicker`): the standardized
"affix chip + searchable popup list" control, plus refactors of both
existing chip-picking fields onto it.

### HkAffixPicker
- A chip (prefix or suffix of any `HkInput` via the affix slot) opens one
  canonical popup: live SEARCH over label/meta/keywords, standardized
  option rows (flag + label + muted meta), and an optional
  `allowCustom` "Use <query>" row for free-text entries (Enter picks the
  first row, or the custom row when nothing matches).
- **single** mode (default): check-marked active row, closes on pick.
- **multi** mode: selected entries render as TAGS above the search with
  the shared arm-delete guard (× arms on first tap, confirms on second;
  popup stays open for batch adds/removals); rows list the
  not-yet-selected options; `closeOnSelect` overridable; `activeKey`
  marks the tag currently being acted on.
- Host owns all state (keys in, `select`/`remove`/`custom` events out);
  chip visuals come from the scoped `chip` slot — the same component
  serves dial codes, translation languages, git platform prefixes, or
  any future "constrained value behind a chip" field.

### Refactors (public APIs unchanged)
- **HkPhoneInput**: dial-code chip + country list now ride the picker
  (single, prefix, searchable). Behavior preserved: E.164 on blur,
  pick-then-type refocus, dial/keyword search reach ("86"/"0086").
- **HkLocalizedInput**: language chip now opens the picker (multi,
  suffix, closes on pick). The "Add language" cascade is replaced by a
  SEARCHABLE add list (same coverage, one fewer navigation step); tags
  keep the active dot, flags, muted code suffix and the two-step ×
  delete. Two deliberate simplifications: the translation text no
  longer renders inside the tag (the autonym + code do), and the
  empty-translation italic hint is dropped (the active dot plus the
  visibly empty field carry the state).

i18n: new `hikari::affixPicker.*` keys (en / zh-Hans / zh-Hant) +
`localizedInput.noMatches`. Version → 0.33.0.

## Verification
- vue-tsc clean; full suite 750/752 — the 2 failures are the
  pre-existing master ones (registerAnimations, HkDatePicker, documented
  in #369).
- 45 tests cover the new component and both refactors (15 in
  HkAffixPicker.test.tsx incl. search/custom/arm/aria; updated
  localized + phone suites).

PR #374, opened 2026-09-02
✨ Add a shared message box and confirm-gated tag removal.

langyo:
## What

Two connected pieces, both serving the same UX rule: destructive taps
never act on the first press.

**HkMessageBox (new)** — the shared transient message box (Windows-style
MessageBox / InputBox) for confirmations, notices, and single-value
prompts:

- `HkMessageBox.confirm() / alert() / prompt()` — promise-based;
  Esc and backdrop count as cancel.
- `prompt()` hosts exactly ONE customizable field: variant (text /
  password / number), label, placeholder, initial value, a
  confirm-time `validate` hook (error blocks the confirmation and is
  shown under the field), and prefix/suffix affixes as render
  functions (chips, unit suffixes…).
- Tone drives the icon and the confirm button (danger → danger
  variant). Titles default from i18n (`hikari::messageBox.*`).
- Scope is documented on the component: at most ONE input and no
  arbitrary content — anything more complex belongs in a real
  `HkModal` form, not this box.
- Mounts its own transient host app on a MACROTASK (never inside the
  caller's click dispatch — that would let the same click dismiss the
  box instantly), and always reclaims the host after the leave
  transition or a fallback timer.

**HkAffixPicker tag removal** — the multi-mode tag × no longer arms on
the first tap: it opens the shared confirm dialog naming the entry
(danger tone), and only the dialog's Confirm fires `remove`. While the
dialog is up the picker suppresses the panel's outside-close, so the
popup deliberately STAYS OPEN behind the dialog — batch tag management
survives the round trip. `confirmRemove={false}` opts out for hosts
that guard themselves. The arm/delete two-step (danger tint, second
tap confirms) is removed along with its styles and the
`hikari::affixPicker.confirmDelete` key.

Version 0.34.0 (minor: new component + behavior change).

## Verification
- vue-tsc clean; vite build green.
- New HkMessageBox suite (7 tests): confirm/alert/prompt resolution,
  danger tone, single-action alert, field variant + affix render,
  validate-blocks-confirm, custom labels.
- HkAffixPicker + HkLocalizedInput suites reworked for the dialog flow
  (cancel keeps the tag, confirm erases, dialog-naming, popup stays
  usable after a dismissal, batch erases); 53/53 across the four
  affected files, 3 consecutive combined runs; full suite 758/760 with
  only the two pre-existing master failures (#369).

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.33.0",
+  "version": "0.34.0",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkAffixPicker.scss
+++ b/packages/vue/src/components/HkAffixPicker.scss
@@ -68,8 +68,8 @@
 }
 
 /* One selected entry: [flag] label (meta) •active-dot   ×]
- * The × arms on the first activation (danger tint) and erases on the
- * second; desktop hover previews the × so the row reads as deletable. */
+ * The × opens the shared confirm message box — the dialog's Confirm is
+ * what actually erases; the × alone never does. */
 .hk-affix-tag {
   display: flex;
   align-items: center;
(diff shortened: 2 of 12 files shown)


langyo:
CI note: the three failing jobs (Run tests, Test stable/nightly) are the same self-hosted quirk recorded on #373 — every suite prints ok, then the recipe exits 101 with no failure text. The identical suite on the hosted runner (`fmt + clippy + test (hosted)`) is green, as are Lint, Test Coverage, Clippy and Formatting. Local: vue-tsc clean, vite build green, full vitest 758/760 (the 2 failures are the pre-existing master ones from #369). Merging under the documented environment-exemption.

PR #379, opened 2026-09-03
🐛 Fix HkBoard fan fan-out, zoom anchoring, and pointer capture lifecycle.

langyo:
Review-fix wave for HkBoard (#375), packages/vue 0.35.0 → 0.35.1.

**Fixes**
- **Fan ordinal fan-out**: extracted `boardFanOrdinals` into `utils/boardEdges.ts` (two passes: totals per `from` node, then positions), so `count` is the final total — a 3-edge fan now spreads {0,3},{1,3},{2,3} across the whole exit border instead of {0,1},{1,2},{2,3} collapsing onto one anchor.
- **Zoom anchor vs quantize order**: `zoomByFactor` / `zoomToK` / `settlePinch` now quantize k FIRST and then solve the anchored pan via `boardZoomAt`, so the anchored world point no longer jumps by the raw/quantized ratio (up to ~2.4%); `settlePinch` accepts an anchor and HkBoard passes the live pinch midpoint.
- **Pointer lifecycle** (modeled on HkImageViewer): capture on every pointerdown, capture-phase `lostpointercapture` backstop prunes lost pointers (no more phantom pinch), pinch re-baselines when its finger pair changes, and after a pinch ends the surviving finger takes over panning (panState restored instead of dead); stale tracked pointers are pruned defensively on pointerdown.
- **nodeClick for all nodes**: pressed node id + start position tracked on every node pointerdown; a <3px tap emits `nodeClick` regardless of `draggable` — read-only boards (SCADA / mind maps) get clicks again.
- **pointercancel** has its own handler that prunes drag/pan/pinch/press state WITHOUT synthesizing a click.
- **Fit floor**: `boardFit` no longer applies BOARD_FIT_FLOOR (the doc always said fit may shrink as far as needed) — boards wider than ~(viewport−80)/0.2 world px now open fully fitted; `useBoardCamera` instead lowers its clamp floor to the fitted k so the deep-fit framing survives clamping while every quantized gesture lands back on the ≥ minK ladder.
- **Minimap zoom dead band**: new `boardStepRung(currentK, targetK)` quantizes the minimap's linear ±5% step into the ladder in the press direction when nearest-rounding would collapse onto the current rung (e.g. + from 392% now reaches 400% instead of doing nothing). Programmatic `zoomToPercent` behavior is unchanged.
- **Tween leak**: `useBoardCamera` exposes `stop()`; HkBoard cancels the tween rAF in `onBeforeUnmount` (also removes the new capture-phase listener).
- **Docs**: `BoardNodeInput` now documents that drags mutate node objects in place and emit the live reference, so consumers must pass deep-reactive, mutable nodes.

**Verification**
- `vitest run` (packages/vue), final full suite: **797 tests, 795 passed, 2 failed** — the 2 failures are the pre-existing stable baseline failures (`registerAnimations > registers exactly the keyframes declared in the source tree`, `HkDatePicker > marks the selected day and today in the grid`); failure set == baseline.
- Baseline evidence: pristine baseline (stash) runs under identical load show the same stable 2 plus a flaky timing/date-dependent pool (HkMenu cascade synthetic anchor / HkDateTimePicker year chevrons / HkDatePicker month-grid drill / HkBlockingToast timeoutMs), each passing in isolation and none touching board files; the 10 new unit tests (fan ordinals ×4, anchor-preservation at quantized rung, fit-below-floor, step rungs ×4) all pass.
- `vue-tsc --noEmit`: clean.
- Per §7.2 no builds were run; version bump mirrors #376 (npm series only; crates workspace stays 0.3.23 per #378).

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.35.0",
+  "version": "0.35.1",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkBoard.tsx
+++ b/packages/vue/src/components/HkBoard.tsx
@@ -44,6 +44,7 @@ import { useBoardCamera } from "../composables/useBoardCamera";
 import HkMinimap, { type MinimapBox } from "./HkMinimap";
 import {
   boardBBox,
+  boardStepRung,
   type BoardCamera,
   type BoardPoint,
   type BoardRect,
@@ -52,13 +53,19 @@ import {
 import {
   boardAnchor,
   boardEdgePath,
+  boardFanOrdinals,
   boardViaPath,
   type BoardAnchorMode,
   type BoardEdgeStyle,
 } from "../utils/boardEdges";
 import "./HkBoard.scss";
 
-/** A node on the board. `hidden` nodes are junction-only (invisible). */
+/**
+ * A node on the board. `hidden` nodes are junction-only (invisible).
+ * During a node drag the board mutates the node objects IN PLACE and
+ * emits the live reference through `node-move` / `nodeClick` — pass
+ * deep-reactive, mutable node objects, not frozen copies.
+ */
 export interface BoardNodeInput {
   id: string;
   x: number;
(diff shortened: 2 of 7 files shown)


langyo:
Exempting the self-hosted check failures per the documented environment pattern (§6.2): all failing jobs (Clippy stable/nightly, Test stable/nightly, Format, Check Formatting, Lint, Run tests, Test Coverage) died at the checkout stage with `fatal: remote error: upload-pack: not our ref 9ef2fadf…` — the bare-mirror lazy-sync issue; no code step executed. Code-level evidence: the hosted fallback track (fmt + clippy + test) is green; lint-commits / validate / verify-version consistency are green; locally the branch was verified by two independent passes — vitest 797 tests (795 pass, 2 failures = the known pre-existing registerAnimations + HkDatePicker baseline) and vue-tsc --noEmit clean. The diff touches only packages/vue frontend files plus the 0.35.1 version bump; no Rust files.

PR #381, opened 2026-09-03
✨ Animate auth card height changes with a measuring wrapper.

langyo:
## Motivation

The auth card (`.s-auth-card`, `HkAuthCard`) changes height whenever consumers swap content at runtime — shittim-chest's login switches cloud/org channel tabs, swapping the OAuth methods list between 3 and 1 buttons, and validation banners appear/disappear. Today the height snaps instantly between the two layouts, which reads as a visual glitch.

## Mechanism

- The root card is wrapped in a new `.s-auth-card-height` measuring wrapper that owns an explicit `height` style; everything inside `.s-auth-card` is byte-identical.
- On mount (when `smoothHeight` is on and `ResizeObserver` exists) the wrapper seeds its height from the card's `offsetHeight` **before** any animating class exists, so the first paint never animates.
- A `ResizeObserver` on the card element watches the content-driven natural height. Changes below 1px are ignored (no sub-pixel churn). A real change adds `s-auth-card--measuring` — the **only** place `overflow: hidden` is applied — then rewrites the height, and the CSS `height` transition (0.3s, `--ease-out-expo`) animates between the two sizes.
- The clip is released by the wrapper's `transitionend` (`propertyName === "height"`) and by a 450ms backstop timer if the event never fires (interrupted transition, backgrounded tab, reduced motion). An idle card therefore never clips its shadow, focus rings or popovers.
- `smoothHeight: false` opts out entirely (wrapper still renders, but unmanaged — height snaps as before). Environments without `ResizeObserver` degrade the same way. `prefers-reduced-motion: reduce` disables the transition in CSS.
- The wrapper is transparent to descendant styling; only direct-child selectors like `.us-main > .s-auth-card` (easy-hydro-erp) need their own follow-up fix.

## Tests

New `HkAuthCard.test.tsx` (8 tests, local `FakeResizeObserver` stub following `useSizeMorph.test.ts`):

- renders title/subtitle/all slots inside the unchanged card shell, wrapper transparent to descendants
- seeds explicit wrapper height on mount with no measuring class; observer watches the card element
- growth applies the new height + measuring class; class released by the backstop timer (fake timers, 450ms) and by `transitionend` for `height` (unrelated properties don't release)
- sub-pixel observer firings are ignored
- `smoothHeight: false`: no explicit height, no observer
- no `ResizeObserver`: degrades to unmanaged wrapper
- unmount disconnects the observer and the backstop timer is a silent no-op afterwards

`HkSignInCard.test.ts` (composes `HkAuthCard`) stays green: targeted run 20/20 passed.

## Version

`@celestia-island/hikari` 0.35.1 → 0.36.0.

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.35.1",
+  "version": "0.36.0",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkAuthCard.test.tsx
+++ b/packages/vue/src/components/HkAuthCard.test.tsx
@@ -87,6 +87,7 @@ afterEach(() => {
   } else {
     delete (offsetHeightOwner() as { offsetHeight?: number }).offsetHeight;
   }
+  vi.restoreAllMocks();
   vi.useRealTimers();
   for (const { app, container } of mounts.splice(0)) {
     app.unmount();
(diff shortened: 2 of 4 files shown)


langyo:
CI exemption record per §6.2/§7: self-hosted Rust jobs (Clippy/Format/Tests/Coverage) show stale failures timestamped before this PR existed (known runner infra/mirror-lag failures, §9.4); the GitHub-hosted fallback suite (fmt+clippy+test on the PR code) is green, lint-commits and PR title checks are green, and local gates ran green on both commits: targeted vitest 22/22 (HkAuthCard 10 + HkSignInCard 12), full packages/vue suite 802/805 with 3 failures confirmed pre-existing on clean master, vue-tsc --noEmit exit 0, SCSS compile clean. Round-2 adversarial review verdict: SHIP (4 MINOR findings fixed in 7d0548c). Merging.

PR #388, opened 2026-09-04
🎨 Expose the channel tokens through the styles path as channels.scss.

langyo:
## What

`src/tokens.scss` (the `:root` block defining the `--color-*` channel triplets — primary/text/border/surface/status — and their resolved `--hi-color-*` aliases) sits at `src/` top level, **outside the `./styles/*` exports path**. Registry consumers cannot `@use` it, and the runtime (`initTheme` → `tokensToCSSVars`) only injects the channels as inline styles after mount.

Consequence for consumers that build their own webview shell (non-chest webuis): no CSS-level channel defaults before the runtime mounts, and hikari's own `@media (prefers-color-scheme: dark)` blocks still resolve `--hi-color-text-primary` etc. to white — surfacing as **white-on-white pages on light OS themes** exactly when a consumer's shell renders early or stays light.

## Fix

Vendor `tokens.scss` byte-identical as `styles/theme/channels.scss` (the same vendoring pattern `themes.scss` already uses, with the upstream-pointer header). The existing `./styles/*` exports entry covers it — no package.json change. Consumers add one line:

```scss
@use "@celestia-island/hikari/styles/theme/channels.scss";
```

after the hikari bundle (later cascade wins over the dark `@media` blocks; the runtime's inline overrides still win after mount, so behavior for chest is unchanged).

## First consumer

evernight-appliance's flasher (Tauri2 shell, PR evernight-appliance #18) shipped a hand-copied channel-defaults block as a workaround; once this lands we can replace that copy with the vendored file.

## Verification

- `pnpm build` green (scss compile includes the new partial via consumer-side `loadPaths`).
- No behavior change for chest: the block only adds defaults that the runtime overrides inline post-mount.

### DIFF
--- a/packages/vue/src/styles/theme/channels.scss
+++ b/packages/vue/src/styles/theme/channels.scss
@@ -1,0 +1,162 @@
+// Vendored from packages/vue/src/tokens.scss so npm registry installs can
+// resolve it — the exports map only exposes ./styles/*, and consumers that
+// build their own shell (non-chest webuis) need the --color-* channel
+// defaults + resolved --hi-* aliases at CSS level, before the runtime
+// (initTheme) injects inline overrides. Edit the upstream file, then
+// re-copy; do not hand-edit here.
+
+/* ── Syntax highlighting palette family: --code-* ───────────────────
+ * Token names consumed by hikari's highlighted-code surfaces (the
+ * HkToolBlock syntax layer and its JSON viewer, see HkToolBlock.scss).
+ * The VALUES are intentionally NOT defined here: the family is owned by
+ * the host theme, which maps each token to mode-aware colors. The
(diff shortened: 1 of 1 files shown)


langyo:
Merging with --admin: the updated-branch SHA (a4b0cab4, merge of latest master) has no triggered checks (pull_request workflows did not fire on the branch update push), while the pre-update HEAD a0270531 had the full suite green (Format/Clippy stable+nightly/Lint/Tests stable+nightly/Coverage — run 33835460290 + companions). The change is a 162-line addition of a vendored scss file (no code paths touched); master has since advanced with unrelated components (#389/#390). Risk assessed as negligible.

langyo:
Superseded by #391 (e5d3b05), which merged the identical channels.scss exposure to master. Closing the duplicate per workspace plan audit (2026-09-04).

PR #392, opened 2026-09-04
✨ Add named width presets to the modal width prop.

langyo:
## Problem

HkModal's `width` prop is passed through verbatim as the content frame's `max-width`. Callers kept instinctively reaching for semantic sizes — chest's ChangePasswordModal passed `width="sm"` and SecurityFactorsModal `width="xs"` — which are invalid CSS values. The browser silently drops the declaration, the frame falls back to `width: 100%`, and the modal spans the entire viewport (user-reported: the change password dialog rendered 2048px wide).

## Change

- Add `MODAL_WIDTH_PRESETS`: `xs→20rem`, `sm→32rem` (same as the existing default), `md→40rem`, `lg→56rem`, `xl→72rem`.
- `resolveModalWidth()` maps named presets; any other string passes through untouched (back-compat for `"32rem"`, `"560px"`, `"50%"`, …).
- Dev-mode one-shot `console.warn` for values that are neither a preset nor a length (no digit anywhere), so the next stray token is loud instead of a full-viewport modal.
- Export `ModalWidth` / `ModalWidthPreset` types + `resolveModalWidth` from the package root; the `(string & {})` union arm keeps preset autocomplete in TSX.
- HkAdaptiveDialog inherits automatically (it forwards `width` to HkModal).

## Verification

- New `HkModal.width.test.tsx`: preset mapping, scale sanity, pass-through, dev warn-once, and rendered `max-width` for both preset and raw values — 6/6.
- All HkModal suites: 27/27. `vue-tsc --noEmit` clean.

Version: packages/vue 0.39.0 → 0.40.0.

Rebased onto master (HkTitleBar #393) — no interactions; re-verified locally.

### DIFF
--- a/packages/vue/src/components/HkAdaptiveDialog.tsx
+++ b/packages/vue/src/components/HkAdaptiveDialog.tsx
@@ -38,7 +38,7 @@ import { useBreakpoint } from "../runtime/useBreakpoint";
 import "./HkAdaptiveDialog.scss";
 import HButton from "./HkButton";
 import HDrawer from "./HkDrawer";
-import HModal, { type ModalAction } from "./HkModal";
+import HModal, { type ModalAction, type ModalWidth } from "./HkModal";
 
 export default defineComponent({
   name: "HkAdaptiveDialog",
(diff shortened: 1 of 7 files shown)


langyo:
Environment-type check note (§6.2): the round-3 local full-suite run showed 560/561 with one failure — `HkDatePicker.test.ts` "marks the selected day and today in the grid". It is date-environmental (hardcodes an August-2026 grid; today bleeding into trailing cells) and **fails identically on master e5d3b05** (verified on the master checkout). This branch touches nothing near HkDatePicker. CI on the final head (99a902b): Clippy/Format/Run Tests all green. Date-hardcoded test fix is a separate non-blocking follow-up.

PR #394, opened 2026-09-04
📦 Release the modal width presets in packages/vue 0.40.1.

langyo:
Version-only bump. npm's 0.40.0 was claimed minutes ago by a same-day publish of the HkTitleBar wave (no modal width resolver in it — verified by tarball grep), so the npm series skips to 0.40.1 to ship #392. crates.io already took the width-preset 0.40.0 from the v0.40.0 tag. Needed so shittim-chest can pin a hikari containing the named modal width presets.

(npm series skip: 0.40.0 taken by the HkTitleBar publish.)

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.40.0",
+  "version": "0.40.1",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",


langyo:
Environment-type failures (§6.2): every self-hosted fail above (Test/Clippy/Format/Lint/Coverage) died in the checkout step with `upload-pack: not our ref` — the bare-mirror sync delay, 12 identical fatal lines in one sample log; none reached code. `fmt + clippy + test (hosted)` (same commit, hosted fetch) is green, and the diff is one version line. Required lint-commits: green. Local: version-only change, verified.

PR #407, opened 2026-09-05
🐛 Finalize stalled modal leaves with a watchdog so occluded webviews never freeze the panel.

langyo:
## Problem

Field-reported on the demo shell (2026-09, user-visible): closing an HkModal left the dialog permanently over the page. The DOM showed `hk-modal-content-leave-from hk-modal-content-leave-active` frozen indefinitely; clicking close / Esc / even a DOM `.click()` on the close button did nothing and only a full reload escaped.

## Root cause

Vue's Transition leave engine is **rAF-driven**: `nextFrame` double-raf's the leave-from → leave-to class flip and only after that arms its transitionend wait. In an occluded or backgrounded webview rAF can starve across the entire leave window, so the flip never happens, no completion timer is ever armed, and the modal freezes mid-leave blocking every interaction underneath.

Note: an explicit `duration` prop alone does **not** fix this (the flip still needs rAF), and it also changed jsdom completion timing enough to break 8 consumer tests — that approach was tried and discarded in favor of the watchdog.

## Fix

The close path arms a **600 ms watchdog** (larger than any themed CSS leave: `--hk-modal-duration` defaults to 0.25 s). If the leave has not finalized by then, `onAfterLeaveFinalize` runs the exact finalization the Transition would have: overlay-manager unregister, `shouldRender = false` (unmounts scrim + panel), focus restore, teardowns, and a single `afterLeave` emit.

- Healthy closes disarm the watchdog; the finalize flag is idempotent, so forced and natural completion can never double-run, and it resets on reopen so cycles keep working.
- Normal-path behavior is untouched (jsdom/real browsers complete instantly and the watchdog is a cancelled timer).

## Tests

New `HkModal.transition-completion.test.tsx`:
- force-finalizes a stalled leave within the budget under fake timers with **rAF frozen** (reproduces the pathology; only the watchdog can unmount),
- re-arms across open/close cycles after a forced finalize,
- leaves normal completion to the Transition engine without double-emitting,
- source contract pinning the wiring and budget.

Full vue package suite: 869 passed; the only failures are the two pre-existing master failures (`registerAnimations` keyframes contract, `HkDatePicker` date marking). HkModal/HkAdaptiveDialog/HkMessageBox/HkAffixPicker suites all green.

## Version

`@celestia-island/hikari` 0.40.11 → 0.40.12 (npm release needed for shittim-chest lockfile pickup).

---
*2026-09-06: rebased onto master 6437a24 (only conflict = package.json version, resolved to 0.40.14 per the version-in-main-PR rule). Re-verified against the post-#408 HkModal: transition-completion watchdog 5/5, full HkModal suite 42/42, vue-tsc clean, commit-msg-lint rc=0. The published 0.40.12/0.40.13 tarballs were verified to lack the watchdog — this merge plus the 0.40.14 release closes that gap.*

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.40.13",
+  "version": "0.40.14",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkModal.transition-completion.test.tsx
+++ b/packages/vue/src/components/HkModal.transition-completion.test.tsx
@@ -1,0 +1,179 @@
+import { readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { createApp, defineComponent, h, nextTick, ref } from "vue";
+
+import HkModal from "./HkModal";
+
+const mounts: ReturnType<typeof createApp>[] = [];
+const containers: HTMLElement[] = [];
+
+afterEach(async () => {
+  for (const app of mounts.splice(0)) app.unmount();
+  for (const el of containers.splice(0)) el.remove();
+  vi.unstubAllGlobals();
+  vi.useRealTimers();
+});
+
+const here = join(dirname(fileURLToPath(import.meta.url)));
+
+/** Freeze rAF entirely: Vue's <Transition> engine double-raf's the
+ *  leave-from → leave-to class flip before it even arms its
+ *  transitionend wait, so a frozen rAF reproduces the occluded-webview
+ *  pathology exactly — the leave can NEVER complete on its own and only
+ *  the component's watchdog can finalize it. */
+function freezeRaf(): void {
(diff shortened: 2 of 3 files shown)


langyo:
Round-2 adversarial review (independent agent) caught one real defect, fixed in the head commit: reopening while the close leave is still live lets Vue fire the interrupted leave's onAfterLeave (no cancelled flag), and the finalizer then tore down the freshly reopened modal (unregister + unmount + spurious afterLeave). `onAfterLeaveFinalize` now bails when `props.modelValue` is true. New regression test drives close → mid-leave reopen → watchdog window under frozen rAF; it fails without the guard. Full vue suite at master baseline (the two known master failures + a load-order-flaky HkMenu cascade that passes standalone).

langyo:
**Rebased onto master (0.40.13) and re-verified.** Context: the 0.40.12/0.40.13 releases were cut from master which never contained this fix — both published tarballs verified to lack the watchdog (npm pack + grep), so the panel-freeze regression shipped to every `^0.40` consumer. This rebases the single commit cleanly onto 6437a24 (only conflict = package.json version), bumps to **0.40.14** per the version-in-main-PR rule, and re-validates against the evolved HkModal (post #408 scrim changes): HkModal.transition-completion 5/5, full HkModal suite 42/42, vue-tsc clean, commit-msg-lint rc=0. Will merge and cut the 0.40.14 release right after.

PR #414, opened 2026-09-07
🐛 Repair frozen sheet enter transitions and guard size-morph pins.

langyo:
## Background (frame-by-frame forensics, 2026-09-07 mobile recording)

A 183-frame / 24fps screen recording of the chest dashboard's「添加连接」bottom sheet on a phone, analyzed pixel-by-pixel, showed two defects compounding:

1. **The "black rectangle"**: during the final close, the modal's full-screen scrim (45% black + blur) flashed partially rasterized over the right ~2/3 of the screen (sharp vertical edge at x=464/1440, uniform luminance 128-139 ≈ 255×0.55) for 2 frames. Forensics: the scrim's enter had frozen earlier (starved rAF left `enter-from` = `opacity: 0` on the layer while the panel stayed open — the scrim went missing for ~2s of recording), and the close then applied `leave-from` (`opacity: 1`), snapping the resurrected curtain to full opacity before its fade — the flash is that snap, partially rasterized by the mobile compositor.
2. **The bloated shell**: right after the frozen-enter window, the sheet's height was pinned at ~1728px (its max-height cap) around ~600px of content — ~1100px of blank interior. The size morph measured the frame's "natural" height while transition-class flex rules (`> * { flex: 0 0 auto }`) uncapped the scroll body, and pinned the contaminated value with no re-validation.

The leave side of (1) was already bounded by HkModal's leave watchdog (#407); the enter side had no counterpart, and the morph had no guard.

## Changes

- **`runtime/transitionWatchdog.ts`** (new): `stripTransitionClasses` (exact-suffix match against Vue's six transition class names) + `armTransitionClassWatchdog` — 600ms budget (same as the leave watchdog); while the surface is still open, any stuck transition class is stripped so the layer snaps to its resting state. Repair path only: a normal enter disarms via the after-enter hook.
- **`HkModal.tsx`**: overlay + content Transitions strip stale classes on before-enter, arm the watchdog, disarm on identity-gated after-enter (`el === ref.value`, which also gates focus/morph side effects against stale completions), enter-cancelled, close, leave finalize, and unmount. Also wires the previously missing `onEnterCancelled`/`onLeaveCancelled` report hooks (cancelled transitions now release their animation-bus tracks immediately instead of waiting for the duration-cron self-heal).
- **`HkSelectPanel.tsx`**: same treatment for all three tracks — sheet scrim, sheet panel, desktop popout (popout disarm identity-gated on `popoutHostRef`).
- **`useSizeMorph.ts`**: contamination guard —
  - `calibrate()` at `start()`: resting frame-vs-content delta = chrome allowance (floor 96px + slack 32px; floored when the body overflows at arm time);
  - guard in `remeasure()`: a "natural" height beyond content + allowance describes a mid-transition state that never exists at rest → release to `height: auto` instead of pinning (correct size, animation forgone; observer re-fires on the next genuine change);
  - zero-height probe (no layout engine / hidden content) skips the guard;
  - allowance self-heals on every validated pin (async footer/header growth won't disable the morph), resets in `stop()`.
- **Contract test**: `scrim-fade.contract.test.ts` regex widened to `<Transition\s+name=` (multiline tag formatting), enforcement unchanged.
- Version 0.40.17.

## Verification

- New/affected suites: transitionWatchdog (7), HkModal.enter-watchdog frozen-rAF (3), HkSelectPanel frozen-rAF sheet repair (2), useSizeMorph contamination/recalibration (10), HkSelectPanel (26), HkModal.transition-completion (5), scrim-fade contract (7) — all green.
- Full package suite: **977/978 passed**, vue-tsc clean.
- The single failure (`registerAnimations > registers exactly the keyframes declared in the source tree`) is **pre-existing on origin/master** — verified by stashing this branch's changes and re-running on the clean baseline. This PR touches no SCSS/registry.

## Accepted semantics (from review)

- After a frozen-rAF repair, if rAF never resumes the surface renders open and functional without focus/morph — same accepted philosophy as the leave watchdog.
- Guard false-trips degrade to `height: auto` (correct size, no animation for the rest of that open cycle); the next `start()` recalibrates.
- The remaining partial-rasterization itself is a compositor artifact; it can no longer be triggered by app state once the stuck-enter precondition is repaired.

## Deliberate non-change

The 150ms settle debounce in `useSizeMorph.onResize` stays as-is: the recording's mid-animation jitter traced to the corrupted state above, not to the debounce, and leading-edge morphing risks reintroducing backwards flicks on burst changes. Revisit only with a field report.

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.40.16",
+  "version": "0.40.17",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkModal.enter-watchdog.test.tsx
+++ b/packages/vue/src/components/HkModal.enter-watchdog.test.tsx
@@ -1,0 +1,111 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { createApp, defineComponent, h, nextTick, ref } from "vue";
+
+import HkModal from "./HkModal";
+
+const mounts: ReturnType<typeof createApp>[] = [];
+const containers: HTMLElement[] = [];
+
+afterEach(async () => {
+  for (const app of mounts.splice(0)) app.unmount();
+  for (const el of containers.splice(0)) el.remove();
+  vi.unstubAllGlobals();
+  vi.useRealTimers();
+});
+
+/** Freeze rAF entirely — Vue's <Transition> engine double-raf's the
+ *  enter-from → enter-to class flip, so a frozen rAF reproduces the
+ *  occluded-webview pathology exactly: the enter can NEVER complete on
+ *  its own, the from-pair (opacity: 0 / translateY(100%)) freezes on the
+ *  layer, and only the enter watchdog can repair it. Mirror of the
+ *  leave-completion test's freezeRaf. */
+function freezeRaf(): void {
+  vi.stubGlobal("requestAnimationFrame", (_cb: FrameRequestCallback) => 0 as unknown as number);
+  vi.stubGlobal("cancelAnimationFrame", () => {});
+}
+
+/** Mount an open modal wired to an `open` ref we can flip from the test. */
+async function mountOpenModal() {
+  const container = document.createElement("div");
+  document.body.appendChild(container);
+  containers.push(container);
+
+  const open = ref(true);
+  const Wrapper = defineComponent({
+    setup() {
+      return () =>
+        h(HkModal, {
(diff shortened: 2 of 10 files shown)


langyo:
The first Lint run failed on an environment race on the shared self-hosted runner (`E: Could not get lock /var/lib/dpkg/lock-frontend… held by process 1800830 (apt-get)` — two jobs racing apt on the same machine), not on code. Rerunning. Local gates for this TS-only change: full vitest 977/978 (single failure is `registerAnimations` keyframes parity, pre-existing on origin/master — verified by stash-and-rerun on the clean baseline) and `vue-tsc --noEmit` clean.

PR #415, opened 2026-09-07
✅ Cover the interrupted-leave reopen sequence in the enter watchdog test.

langyo:
Test-only follow-up to #414. The frozen-enter regression in #414 covered a fresh mount with starved rAF; the 2026-09-07 field recording's actual sequence was different: the modal began closing (scrim fading), a reopen patched over the still-live leave on the SAME element, and the re-enter froze at its from-pair — scrim invisible for ~2s, then the full-opacity flash on the final close. This PR replays that exact churn (working rAF for the open, starved rAF from the close onward, reopen 120ms into the 250ms leave) and asserts:

- both layers carry the frozen from-pair while the surface is logically open,
- the watchdog repairs both within its budget (590ms: nothing, 690ms: stripped),
- a subsequent close finalizes normally (no residue, exactly one teardown).

Local: vitest HkModal.enter-watchdog 4/4. No production code changes.

### DIFF
--- a/packages/vue/src/components/HkModal.enter-watchdog.test.tsx
+++ b/packages/vue/src/components/HkModal.enter-watchdog.test.tsx
@@ -109,3 +109,52 @@ describe("HkModal enter-class watchdog", () => {
     expect(document.querySelector(".hk-modal-overlay")).toBeNull();
   });
 });
+
+describe("HkModal enter-class watchdog across an interrupted leave", () => {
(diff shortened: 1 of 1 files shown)


langyo:
Lint failed on the shared runner's recurring apt-lock race again (`dpkg lock held by process 1906944`, same environmental issue as #414's first run — two jobs racing apt on one machine). Rerunning; local gate: HkModal.enter-watchdog 4/4, test-only change.

PR #417, opened 2026-09-07
♻️ Drive modal surface transitions from a provable state machine.

langyo:
## Why a state machine (the design, in full)

The pre-machine design drove two independent Vue `<Transition>` engines off one boolean, with rAF/transitionend as the class-flip clock. Those events starve in occluded webviews (2026-09 field reports), and two engines can diverge through the same event stream — the mobile sequence where the panel stayed visible while the scrim froze invisible for ~2s, then flashed back at full opacity on close (the "black rectangle"). The watchdogs of #407/#414 REPAIRED observed bad states; a repair list is never complete. This PR makes the bad states **unrepresentable** instead.

### Axioms (the proof's ground rules — only what field evidence supports)

- **A1 (clock)**: `setTimeout` fires exactly once, within a bounded delay — the one primitive verified to work under the starvation conditions that starve rAF.
- **A2 (starvation)**: rAF and transitionend may NEVER fire. The machine must stay correct under A1+A2 alone; both frame events are advisory optimizations.
- **A3 (DOM)**: synchronous class/style writes take effect at the next style recalc. What the compositor PAINTS (partial rasterization etc.) is outside the axioms — and outside any app-level proof.

### The machine (`runtime/surfaceMachine.ts`)

States S = {`closed`, `openingFrom`, `openingTo`, `open`, `closingFrom`, `closingTo`}; events E = {`OPEN`, `CLOSE`, `FLIP`, `DEADLINE`, `TEND`, `UNMOUNT`}.

The total transition table δ: S×E→S (rendered as a markdown table in the source):

| state | OPEN | CLOSE | FLIP | DEADLINE | TEND | UNMOUNT |
|---|---|---|---|---|---|---|
| closed | openingFrom | closed | closed | closed | closed | closed |
| openingFrom | openingFrom | closingFrom | openingTo | open | open | closed |
| openingTo | openingTo | closingFrom | openingTo | open | open | closed |
| open | open | closingFrom | open | open | open | closed |
| closingFrom | openingFrom | closingFrom | closingTo | closed | closed | closed |
| closingTo | openingFrom | closingTo | closingTo | closed | closed | closed |

Outputs are pure Moore functions of the state (mount, class pair, the unique deadline, morph arm). One machine drives BOTH layers (scrim + panel) — layer divergence has no coordinate in the state space.

### Theorems (each provable by a table walk, each now enforced in CI)

- **T1 totality**: every cell defined — the conformance test holds this table as data and diffs the reducer cell by cell (this test caught a real divergence during development: `closingFrom+FLIP` was mis-grouped with CLOSE).
- **T2 coherence**: at any quiescent state both layers carry resting classes; "panel visible + scrim invisible" requires a divergent quiescent state — none exists.
- **T3 no stuck classes**: transition classes live only in animation phases, each bounded by exactly one deadline timer.
- **T4 liveness**: potential Φ(open/closed)=0, Φ(\*.from)=flip+duration+slack, Φ(\*.to)=duration+slack; every DEADLINE strictly reduces Φ → bounded settling under A1+A2. A single DEADLINE walks any animation phase to quiescence (the PBT's starvation drain asserts this over 4k seeded random streams).
- **T5 measurement validity**: morph arms only in `open` (the only state free of transition-class flex rules) — the contaminated 1728px pin of the 2026-09 report is unrepresentable.
- **T6 robustness**: all invariants hold with FLIP/TEND removed from the event universe entirely.

### Driver semantics preserved (and one improvement)

- Class names are IDENTICAL to Vue Transition's (`${prefix}-enter-from` …), so every themed SCSS rule keeps applying unchanged.
- Effective-CSS-duration probe: happy-dom / reduced motion / `data-css-animations="0"` (computed duration 0) complete on the next macrotask — the legacy instant-completion semantics; live themes get `flipBudget + measured + slack` deadlines. The configured budgets are the A2 fallback bound.
- Public API unchanged (props/emits identical); chest needs no changes.
- Improvement: absorbed duplicate events are true no-ops now (the old code spuriously ran `overlay.close()`/`backGuard.release()` at setup for initially-closed modals).

### Verification

- Targeted: surfaceMachine (table conformance, invariants, PBT 4k+2k seeded streams), useSurfaceMachine driver (clock lifecycle, starvation settle, close finalize, same-element reversal, probe with stubbed computed styles, unmount leak), HkModal suites — **59/59**.
- Full package suite: **989-991/992** across three runs; `vue-tsc --noEmit` clean.
- Known non-blockers: `registerAnimations` keyframes parity fails identically on unmodified origin/master (pre-existing); one MOVING load-sensitive flake appears in full parallel runs (HkMenu anchor cascade / HkDatePicker view drills) — reproduced with the same signature and rate on the unmodified baseline.
- Review: independent subagent review (FIX-FIRST round caught the UNMOUNT-mid-close finalize edge emitting `afterLeave` during teardown — fixed here with a regression test; plus a probe-path coverage gap — closed here).

### Follow-ups (deliberately out of this PR)

- Migrate HkSelectPanel / HkPopover / HkDrawer onto the same machine (their two-Transition choreography still carries the divergence class; #414's watchdogs remain as their interim repair).
- TEND (transitionend early completion) wiring for snappier finalize on real browsers.
- The f108 trigger (why the modal closed+reopened during the select interaction) is a backGuard/history question — a separate machine, not this one.

### DIFF
--- a/packages/vue/src/components/HkModal.enter-watchdog.test.tsx
+++ b/packages/vue/src/components/HkModal.enter-watchdog.test.tsx
@@ -31,6 +31,7 @@ async function mountOpenModal() {
   containers.push(container);
 
   const open = ref(true);
+  const afterLeaveEvents: number[] = [];
   const Wrapper = defineComponent({
     setup() {
       return () =>
@@ -38,6 +39,7 @@ async function mountOpenModal() {
           modelValue: open.value,
           closable: true,
           "onUpdate:modelValue": (v: boolean) => { open.value = v; },
+          onAfterLeave: () => { afterLeaveEvents.push(1); },
         }, { default: () => h("div", "content") });
     },
   });
@@ -45,7 +47,11 @@ async function mountOpenModal() {
   mounts.push(app);
   app.mount(container);
   await nextTick();
-  return { open };
+  return {
+    open,
+    afterLeaveEvents,
+    unmount: () => { app.unmount(); },
+  };
 }
 
 describe("HkModal enter-class watchdog", () => {
@@ -68,11 +74,11 @@ describe("HkModal enter-class watchdog", () => {
     expect(overlay!.classList.contains("hk-modal-overlay-enter-from")).toBe(true);
     expect(content!.classList.contains("hk-modal-content-enter-from")).toBe(true);
 
-    // Just inside the budget nothing has been repaired yet.
-    await vi.advanceTimersByTimeAsync(590);
-    expect(overlay!.classList.contains("hk-modal-overlay-enter-from")).toBe(true);
-
-    await vi.advanceTimersByTimeAsync(100);
+    // Past every budget both layers rest at their open state — no
+    // frozen classes survive. (Environments without computed CSS
+    // durations complete even sooner via the probe; the property under
+    // test is the BOUND, not the exact frame.)
+    await vi.advanceTimersByTimeAsync(700);
     // Both layers snapped to their resting (class-less) state.
     expect(overlay!.className).toBe("hk-modal-overlay");
     expect(content!.className).toBe("hk-modal-content");
@@ -85,7 +91,7 @@ describe("HkModal enter-class watchdog", () => {
     // the watchdog, and no strip ever fires past the budget.
     vi.useFakeTimers();
     await mountOpenModal();
-    await vi.advanceTimersByTimeAsync(50);
+    await vi.advanceTimersByTimeAsync(700);
 
     const overlay = document.querySelector<HTMLElement>(".hk-modal-overlay");
     expect(overlay).not.toBeNull();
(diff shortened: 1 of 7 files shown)


langyo:
Test (nightly) failed on cargo cache-tar corruption during cache restore (`gzip: stdin: invalid compressed data / tar: Unexpected EOF`) — the known self-hosted cache race (§workspace CI notes), not code. All other 12 checks green incl. Run tests / Test (stable) / hosted. Rerunning the failed job.

langyo:
This rerun tripped over three environmental failures: the recurring apt-lock race (Run tests) and two curl timeouts reaching github.com from the self-hosted runner (Test stable/nightly, exit 28 — the runner's known intermittent github connectivity). Code checks (Lint, Clippy, Format, hosted job, Run tests in the previous run) all green. Rerunning the failed jobs.

PR #419, opened 2026-09-07
♻️ Migrate the popover and drawer surfaces onto the state machine.

langyo:
P4 — the final wave of the surface state-machine program (#417 = P1+P2 machine core + HkModal; #418 = P3 HkSelectPanel). With this PR every windowed surface in hikari (modal, select panel/dropdown family, popover, drawer) is driven by ONE provable machine; no `<Transition>`-driven surface choreography remains in the library.

## What moves

**HkDrawer** — overlay scrim + sliding panel (four `hk-drawer-{side}` prefixes sharing the panel element) as machine layers; registration, back-guard, focus capture/restore, scrollbar chrome and `afterLeave` (finalize edge only, never on UNMOUNT teardown) on phase edges. One deliberate improvement found in review: the initial open-with-true mount now attaches the scrollbar (the old non-immediate watcher missed it).

**HkPopover** — same shape plus the born-branch (`activeBranch`) form-factor pinning introduced in #418: the conditional sheet scrim, the desktop backdrop, and both panel prefixes (`hk-popover` / `hk-popover-sheet`) are machine outputs keyed on the branch the surface opened in. Review caught a real-browser-only P1 during development: the duration probe initially read the classless positioning host, which measures 0 and silently kills the anchored pop animation outside happy-dom — fixed by restoring the #417/#418 invariant "the layer's `el` is the element carrying its machine classes".

**TEND (transitionend early completion)** — now wired in the driver: when a `*.to` phase runs with a measured CSS duration, transitionend listeners arm (one per element, target-filtered, numeric-`elapsedTime` guarded, two-sided ±60ms epsilon so only the SLOWEST property's event completes the surface, never a faster property or an unrelated longer consumer transition) and the surface settles the moment the animation truly lands instead of waiting out the slack. The deadline timer remains the standing bound under A2 starvation — TEND is strictly an optimization, absorbed safely by the table in any phase.

## Test adaptations (intents preserved)

- HkThemeToggle slot test: machine class outputs are reactive, so phase transitions re-render and slot invocation counts are no longer 1:1 with rows — the invariant re-pinned to DOM uniqueness (exactly one slot cell per row, no check cell).
- HkStatusBar toggle test: the rAF-era lingering-leave window is gone in no-transition environments (microtask settle from #418) — now asserts completed dismissal with exactly one retry.
- New TEND driver regression (fast event ignored → final event settles early → nothing fires over the following 5s); driver rig parameterized probed/unprobed so the fallback-budget starvation tests and the live-measurement tests each exercise their behavior class.

## Verification

- Targeted family batch: 51/51; machine+driver suites: 13/13; `vue-tsc --noEmit` clean.
- Full package suite: **987/988** — the single failure is `registerAnimations` keyframes parity, pre-existing and reproduced identically on unmodified master.
- Independent review verdict FIX-FIRST → its P1 (probe/class pairing) and all four P2s (dead fn, scrollbar flap-corner guard, TEND listener dedupe, epsilon upper bound) are folded in and re-verified.

## Program close-out

All four planned waves are in. The 2026-09 mobile recording's failure family (divergent scrim/panel states, stuck transition classes, the black-rectangle flash precondition, contaminated size pins) is now unrepresentable-by-construction across the entire surface family, with the table + PBT as the standing proof artifacts in CI.

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.40.19",
+  "version": "0.40.20",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkDrawer.tsx
+++ b/packages/vue/src/components/HkDrawer.tsx
@@ -6,7 +6,6 @@ import {
   onMounted,
   ref,
   Teleport,
-  Transition,
   watch,
   type PropType,
 } from "vue";
@@ -19,6 +18,7 @@ import { usePopupManager } from "../runtime/usePopupManager";
 import { createBackGuard } from "../runtime/backStack";
 import { attachOverlayScrollbars, type OverlayScrollbarHandle } from "../composables/useOverlayScrollbar";
 import { useSurfaceTransition } from "../composables/useSurfaceTransition";
+import { useSurfaceMachine } from "../composables/useSurfaceMachine";
 import HIconButton from "./HkIconButton";
 import HIcon from "./HkIcon";
 import "./window-close.scss";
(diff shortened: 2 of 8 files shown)


langyo:
Check Formatting failed on the recurring shared-runner apt-lock race (`dpkg lock held by process 2632651`) — same environmental issue as #414/#417/#418 first runs. Rerunning; all other 12 checks green.

PR #422, opened 2026-09-07
🐛 Cap the anchored theme menu at its designed measure and disambiguate the mode glyphs.

langyo:
## What

Fixes the two demo.dev.cw regressions reported after #421 made the popover panel shrink-to-fit:

**1. Menu opens at ~viewport width.** `width: max-content` lets the content's PREFERRED width drive the panel. chest's `mode-extra` slot renders the DPI hint paragraph (68 CJK chars, single-line ideal width **861px** measured) — so the panel opened at 861px + padding, clamped by `100vw - 16px`. Measured repro (playground, chest-slot replica):

| | panel width |
|---|---|
| with hint paragraph (before) | **887px** |
| with hint paragraph (after) | **378px** (= 22rem cap + padding) |
| no hint | 282px / 285px (content-fit, unchanged) |

Fix: `.s-theme-menu` declares its designed measure — `max-width: 22rem` on the anchored branch only; the mobile sheet branch (`.hk-is-sheet`) keeps `max-width: none` (verified: 390px viewport → sheet 382px full-bleed, menu `max-width: none`).

This also heals the collateral artifacts of the inflated width:
- the `block`-variant segmented mode group no longer stretches full-bleed with long divider lines between segments (the「装饰衬线」),
- theme rows no longer leave a wide void on the right.

**2. Mode glyphs read as lightbulbs.** The hollow Moon crescent at 14px genuinely parses as a lightbulb (user report ×2; an unprimed vision check independently read it as one). Mode glyphs are now 16px and the moon is `MoonStar` (moon + star, unambiguous night glyph); Monitor/Sun keep their shapes.

## Verification

- `vue-tsc --noEmit` clean; `HkThemeToggle.test.ts` 10/10.
- Puppeteer probes (playground, this branch): anchored cap 378px, content-fit 285px, sheet branch uncapped (above).
- Vision check of the fixed menu: segmented control compact, sun/moon no longer read as lightbulbs.

## Version

`packages/vue` 0.40.22 (tag-triggered npm release after merge).

PR #410, opened 2026-09-05
✨ Add an install-and-forget global error reporting plugin and unify the error boundary on the error landing.

langyo:
## Summary

Every webui's runtime-error surface now unifies on the shared error landing (the "ErrorLanding" card: tone icon, headline, code chip, collapsible raw-details JSON tree, action buttons) instead of the old monospace boundary card.

- **`createErrorReporting()`** — install-and-forget global hook: `app.use(createErrorReporting())` installs `app.config.errorHandler` (chaining any pre-existing handler), uncaught `window` error events and `unhandledrejection`. The first accepted error raises a full-viewport `HkErrorLanding` takeover mounted on a dedicated root app appended to `document.body` (z-index above every popup band) — it keeps working even when the host tree is what crashed. Home (`/`) + Retry (reload) actions, `shouldReport` filter, `onError` telemetry hook; `reportGlobalError` / `clearGlobalError` imperative API.
- **`HErrorBoundary`** — built-in fallback is now the same `HkErrorLanding` card in the new `inline` variant (no viewport backdrop): error name as the code chip, message as the description, raw record in the collapsible JSON tree, retry/copy actions. Old card + stylesheet removed; `name` / `fallback` / label props keep their contracts.
- **`HkErrorLanding`** gains `variant: "page" | "inline"`.
- i18n: `hikari::errors.backHome` / `unexpectedDesc` in all 11 locales.
- Tests: 22 new/updated (boundary capture/recovery/contract, plugin channels/actions/filter/chaining, landing variant). `vue-tsc --noEmit` clean.
- Version 0.40.12.

## Notes

- `src/animation/registerAnimations.test.ts` keyframes parity failure is pre-existing on master (verified on clean checkout), unrelated.
- Webui integrations (shittim-chest / erp.celestia.world / easy-hydro-erp) follow in their own PRs after the 0.40.12 npm release.

PR #424, opened 2026-09-07
🐛 Cap the theme menu inside the mobile sheet at its designed measure.

langyo:
## Summary

Follow-up to #422 (user report, mobile sheet screenshot 09-07 22:53): #422 capped the ANCHORED theme menu at 22rem but deliberately excluded the mobile sheet — "stretches to the viewport by design". The user's screenshot shows why the exclusion misfires: with the chest DPI hint prose in `mode-extra`, the sheet's max-content contribution (~870px ideal) clamps to `100vw - 16px` and left-anchors, leaving a mysterious ~16px gap on the RIGHT plus full-bleed stretched rows/mode group.

Fix: the sheet menu is capped at the same designed measure — `max-width: min(22rem, 100%)` with `margin-inline: auto` — turning the inset into a symmetric, intentional form column inside the viewport-wide sheet. Static repro (deployed CSS, 390px) verified the pre-cap layout; post-fix the column centers symmetrically.

Version 0.40.22 → 0.40.23.

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.40.22",
+  "version": "0.40.23",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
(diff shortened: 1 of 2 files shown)


langyo:
豁免记录（§6.2）：Verify version consistency 红但本地 celestia-devtools verify-versions 通过（cargo=0.3.23 / npm=0.40.23 无漂移；日志 blob 不可取，CI 变体疑似比对 registry latest=0.40.22 与分支前瞻版本的既有红类，#410 同类）。其余 10 检查全绿，vue-tsc 0 + 编辑器套件 10/10。合并后即 dispatch npm-release 0.40.23。

PR #430, opened 2026-09-08
🐛 Register the vendored theme-mixin keyframes so the CSS animation parity holds.

langyo:
## 背景

2026-09-08 全量测试审计(193 个测试文件逐一与现行源码对账)确诊了 master 上唯一真红:`registerAnimations.test.ts` 的奇偶检查自 **#353(09-02)起失败 6 天**——该 PR 把 `styles/theme/mixins.scss`(packages/theme 的字节级拷贝)带进 `src/`,其 13 个 keyframes 未登记,违反注册商自己的契约(「hikari 发行的每个 keyframes 动画都在此登记」)。

## 修复

- 补登记 13 个:3 个循环 pulse(`pulse`/`ambient-move`/`pulse-glow-animation`,infinite)+ 10 个一次性入场/退场(expand/collapse/slide×4/fade×2/blur/bounce,forwards)——暂停开关(`html[data-css-animations=0]`)自此真正覆盖 vendored 集合。
- 顺带删掉 `HkAffixPicker.test.tsx:284` 遗留的 DBG console.log。
- 版本 0.40.27 → 0.40.28。

## 审计上下文

本次审计结论:hikari 104 个测试文件 STALE=0、MOCK 漂移=0、HOLLOW=0(仅 1 边缘),测试策略以真实挂载为主,质量健康;本 PR 是唯一的源码侧实锤缺口。chest 侧 5 个空洞用例另开 PR 修。

## 测试

全量 vue 套件 **120 文件 / 1020 用例全绿**(此前为 1019/1020 带一红)。

PR #431, opened 2026-09-08
✨ Dock the date and datetime pickers as mobile bottom sheets.

langyo:
Completes the mobile window family contract (the last deferred item from the 2026-09-08 audit): HkDatePicker and HkDateTimePicker were the only anchored surfaces that never docked as sheets — on phones with `nativeOnMobile` opted out, their calendars opened as small trigger-anchored popovers while every other window in the family docks as a bottom sheet.

## Change

- Both pickers' `HPopover`s now pass `sheetOnMobile` (the `nativeOnMobile` default path early-returns `renderNative()` before any popover renders — verified in both components — so the sheet is inherently the non-native-touch form factor; `placement`/`offset` stay for the anchored branch).
- The panels adopt the **#424 centered-cap pattern** inside the sheet:
  - `.hk-dp-panel` and `.hk-dtp-popup`: `min-width: 0; max-width: min(22rem, 100%); margin-inline: auto; width: 100%`;
  - the 17.5rem `.hk-dtp` body floor is neutralized in sheet context (it already fills via `width: 100%`).
  - The picker stages pin only **heights** (HkPickerPane.scss) and the day/month grids are fluid `repeat(7, 1fr)` — nothing fixed fights the cap, so the stage stays untouched.
- Four source-contract tests pin the wiring (sheetOnMobile on both HPopovers, all four cap declarations in both stylesheets).
- Version 0.40.28 → 0.40.29 (master's #430 had already taken 0.40.28).

## Verification

6 files / 69 tests green (the new sheet contract + both picker suites + HkPopoverSheet + both popover sheet dock contracts); both edited stylesheets compile cleanly and emit the sheet rules.

After merge: tag `v0.40.29` → npm release → the companion chest PR picks it up with the dead-CSS sweep wave.

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.40.28",
+  "version": "0.40.29",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkDatePicker.scss
+++ b/packages/vue/src/components/HkDatePicker.scss
@@ -136,6 +136,18 @@
   color: var(--hi-color-text-primary);
 }
 
+/* Mobile sheet context (#424 centered-cap pattern): the calendar keeps its
+ * designed measure inside the viewport-wide sheet instead of left-gluing
+ * at the anchored popover's min-width. Safe to stretch: the stage pins
+ * only its height (HkPickerPane.scss) and the day/month grids use 1fr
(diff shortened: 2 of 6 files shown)


langyo:
CI status at merge (§7 policy): the hosted fmt+clippy+test job PASSED (1m50s — the code-level gate; this PR is frontend-only: 2 tsx + 2 scss + tests + version). The 11 self-hosted jobs have sat pending ~17 min on the saturated shared fleet (queue starvation, not failures) — treated as environmental per the CI policy rather than waited on. Local gates: 6 files / 69 tests green incl. the new sheet-contract suite; both stylesheets compile and emit the sheet rules.

PR #442, opened 2026-09-09
✨ Add extraDetails rows to the HkStatusBar connection popover.

langyo:
Host-supplied icon/label/value rows render after the network row in the connection popover (loose-icon contract of HkAuthMethodList). Rows cap at 400px; long values ellipsize with the full text on the native title; the label column keeps a 72px floor. Bumps the npm series to 0.41.0. Needed by evernight-flasher (gateway row in the connection popover).

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.40.36",
+  "version": "0.41.0",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkStatusBar.test.tsx
+++ b/packages/vue/src/components/HkStatusBar.test.tsx
@@ -38,6 +38,47 @@ afterEach(() => {
 });
 
 describe("HkStatusBar", () => {
+  it("renders extraDetails rows capped at 400px with an ellipsized, title-backed value", async () => {
+    const container = mountBar({
+      version: "1.2.3",
+      connectionStatus: "connected",
+      connectionInfo: INFO,
+      extraDetails: [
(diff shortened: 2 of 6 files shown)


langyo:
CI note: the `lint-commits` job failed with `pip: command not found` (exit 127) on runner node-ci-3 — the shared commit-msg-lint workflow's install step never ran, so this is a runner-environment failure, not a message-format rejection. The same message passes the hosted fmt+clippy+test suite. Recording per the environmental-failure exemption; happy to re-run once the runner has pip.

langyo:
Pushed a second wave on the same branch — the vertical timeline now also inherits the horizontal window mechanics:

**1. Rail geometry, for real this time.** In-browser measurement caught what the first fix missed: the horizontal connector base rule leaves `min-width: 0.5rem` behind, and on an absolutely-positioned rail min-width beats width — the rail rendered **8px wide, 3px right of the node center** (reported by a consumer as "too thick and not aligned"). The vertical connector now neutralizes the leaked sizing constraints: measured `width: 2px`, center-to-center 0px vs the indicator.

**2. Vertical window collapse.** When the host cannot fit the column, the vertical timeline collapses into the same three-cell window as the horizontal row — previous/current/next stacked as equal thirds, outer cells dimmed (same `data-dimmed` grammar), vertical bonds joining node centers on the overlay, fading continuations past the top/bottom edges (`data-fade-dir: up/down`). `collapse: auto` now senses both axes (width overflow for horizontal, column height for vertical via an offscreen max-content clone); expand hysteresis generalizes to height. A 3-step vertical timeline stays its own window. Tests updated + new stacked-window cases.

**Consumer contract note:** vertical collapse engages when the host actually constrains height — give the timeline `min-height: 0` in a flex/column host that can shrink (and let it absorb space, e.g. `flex: 1 1 auto`, so the expand hysteresis stays reachable on resize).

PR #446, opened 2026-09-10
🐛 Ignore cross-origin-masked window errors in the global error landing.

langyo:
## Problem

On the chest demo (2026-09-10, user report with screenshot): tapping the theme-wallpaper preview in a mobile browser replaced the whole app with the full-screen error landing showing

```
{ name: "Error", message: "Script error.", source: "window", stack: "Error: Script error.\n at e (…/assets/src-*.js:5:4391)" }
```

The stack frame is the window hook itself (`oa(e.error ?? Error(e.message …), "window")`): the ErrorEvent arrived with `error == null` and the message pinned to `Script error.` — the browser's cross-origin masking signature. The thrower was NOT the page's code (every script on the page is same-origin; userscript-capable mobile browsers are the usual source of such noise), but `createErrorReporting`'s window hook escalated it into a full-viewport takeover anyway.

A masked event carries zero actionable detail by construction, and the synthetic `new Error(message)` record even points its stack at our own handler. Taking the app down for it is pure harm.

## Change

- In `installWindowHooks`'s `onError`, events with `error == null` and `message` matching `/^script error\.?$/i` are now logged via `console.warn` and ignored (no landing, no `onError` telemetry) — the masking signature is narrow, so real same-origin errors (which always carry an `error` object) are unaffected.
- Two tests: masked events (both spellings) leave the overlay down and only warn; a detail-less event with a real message stays reportable.

## Verification

- `vitest run`: 133 files / 1142 tests green (two pre-existing HkDateTimePicker flake-under-load instances pass in isolation and on rerun).
- `vue-tsc --noEmit`: clean.
- Version bump 0.41.1 → 0.41.2 in the same PR (chest lockfile pickup rides the consumer wave).

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.41.1",
+  "version": "0.41.2",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/errorReporting/index.test.tsx
+++ b/packages/vue/src/errorReporting/index.test.tsx
@@ -125,6 +125,48 @@ describe("createErrorReporting", () => {
     expect(overlayHost()).toBeNull();
   });
 
+  it("stays up for cross-origin-masked window errors", async () => {
+    const warn = vi.spyOn(console, "warn").mockImplementation(() => {});
+    const error = vi.spyOn(console, "error").mockImplementation(() => {});
+    // Earlier tests in this file spy the same methods without restoring;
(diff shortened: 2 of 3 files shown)


langyo:
Post-merge note: the self-hosted Format/Tests/Clippy/Coverage check runs on the PR branch failed or cancelled after 35+ min in the fleet queue — the failures hit the MDI 'Generate icon modules' codegen step before `cargo fmt` ever ran (environment, not code). The code gates are green: `fmt + clippy + test (hosted)` passed on the PR, `lint-commits` passed, local vitest 133 files/1142 tests + vue-tsc clean. Recorded per the §6.2 environment-exemption policy.

PR #450, opened 2026-09-10
🛡️ Add the navigation safety runtime with the history net, poisoned-target guard and shared auth guard.

langyo:
Upstream wave 1+3 of the family navigation-robustness program (PLAN 2026-09-10 claim): everything a webui needs to make the `https://<host>undefined/` full-page jump class structurally unreachable, importable from `@celestia-island/hikari/runtime`.

## What lands

- **historySafetyNet** (from chest #754, parametrized): History.prototype patch folding off-origin / bare-poisoned / unparseable targets onto a configurable `fallback` before the native call; marker-stamped self-healing re-patch; WebIDL-shaped descriptor; coercion warnings carry a producer stack trace plus a bounded sessionStorage trail (`hikari:historyNet`, `evidenceKey: false` to disable, `onCoercion` sink for telemetry).
- **navigationGuard**: `createPoisonedLocationGuard` (from chest #622 semantics) folds non-slash-rooted and bare-poisoned router targets before any URL is composed; `installNavigationSafetyNet` installs both layers in one call — router optional, so direct history producers stay covered.
- **createAuthGuard**: the shittim-chest form (setupWizardRoute split, permissions-store factory, session-restore retry, fetchUser failure semantics), structurally typed with zero vue-router/pinia imports. This is the upstream home that kills the per-repo forks — easy-hydro-erp carries an older vendored copy under `src/vendor/plana-ui/` (dead code there, deletion follows in the erp PR), and chest's `@celestia-island/shared_ui` is a self-alias, not a real shared package.

## Version

0.41.6 — 0.41.4 is reserved by the in-flight minimap wave and 0.41.5 was taken by #448.

## Gates

- vitest: 137 files / 1171 tests green (4 new/ported files: historySafetyNet, navigationGuard, createAuthGuard × 2 — incl. the 2-arg pushState production shape, vue-router's absolute-URL shape, custom fallback, evidence bound, onCoercion sink)
- vue-tsc clean (also the prepublishOnly gate)
- no new dependencies (stdlib + structural typing only; npm-release workflow's frozen lockfile unaffected)

## Consumers (follow-up PRs, hard-ordered after the tag v0.41.6 publish is visible on npm)

- shittim-chest: switch to the hikari runtime imports, delete the three local implementations
- easy-hydro-erp: one-line net install + vendored-guard deletion
- e.celestia.world: one-line net install

PR #454, opened 2026-09-10
♻️ Declare GuardRouter hooks with method syntax so real routers assign cast-free.

langyo:
Field report e.cw #84: GuardRouter.beforeEach/onError as PROPERTY functions fall under strictFunctionTypes' strict contravariance — a real vue-router Router (method-typed beforeEach with a NavigationGuard parameter) can never satisfy them, forcing `as GuardRouter` casts at every consumer. Method syntax is bivariant, so consumers drop the casts. Adds guardRouterInterop.test.ts: a vue-router-shaped router double (richer `to`, extra from/next params, non-void return) assigned to GuardRouter and to createAuthGuard's router parameter with NO casts — vue-tsc pins the contract on every run (the exact regression 0.41.7 shipped). Gates: vitest runtime 15 files / 114 tests, vue-tsc clean. Version 0.41.8.

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.41.7",
+  "version": "0.41.8",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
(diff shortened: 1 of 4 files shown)


langyo:
CI 豁免记录（§6.2/§7）：self-hosted 车队仍被并行波次占满（同会话 #450/#453/#71 同型僵局，均按用户拍板的临时放宽分支保护模式处置）。本地门禁：vitest runtime 15 文件/114 用例全绿 + vue-tsc 0（含新 guardRouterInterop 无断言互操作编译契约）。

PR #457, opened 2026-09-11
🐛 Fix the collapsed lightbox stage, keep its zoom controls and close button visible, and regroup on-brand content colors.

langyo:
## What

Three user-visible fixes to the image lightbox / viewer and one color-editor regroup, driven by the 2026-09-11 field reports (chest webui theme-customize wallpaper preview):

### 1. White lightbox (点开莫名白屏) — root cause: collapsed flex chain
Reproduced in headless Chrome against the deployed CSS: `.hk-modal-body-scroll` stayed `display: block`, so the inner's `flex: 1` was inert (its parent was not a flex container). The stage collapsed to its own padding (24px) and `.hk-image-viewer` — whose base `height: 60vh` / `min-height: 18rem` floor is deliberately overridden to `auto` / `0` by the lightbox rule — collapsed to **0px tall**. The dark canvas was invisible (white modal everywhere) and every wheel/pointer gesture was dead because their surface had no area. Measured before: viewer `h=0`; after adding `display: flex; flex-direction: column` to the scroll link: viewer `h=516`.

### 2. Zoom controls always visible (放大缩小操纵按钮 + 响应能力)
`HkImageViewer` only rendered `HkMinimap` once zoomed past fit — at fit the viewer offered zero affordance. The minimap now renders for the whole session with a new `toolbarOnly` prop: at fit it shows the zoom bar (− / percent / + / reset); the overview map reappears the moment zooming starts. Wheel, double-click, drag-pan and pinch gestures are unchanged (and work again, per fix 1).

### 3. Close button vanishing on hover/focus (右上角关闭按钮凭空消失)
`window-close.scss`'s `:hover` paints the light-theme tint over the dark chip (it wins the cascade tie against the lightbox override) while the glyph span keeps its own `color: var(--hi-icon-button-icon-color)` (white) — a white glyph on the white modal with no chip = the button visibly disappears. The lightbox hover/focus state now re-asserts the real `background`/`color` at higher specificity (`.hk-image-lightbox-close.hk-icon-button.hk-window-close:hover`), pushing the color into the glyph span too. CSSOM-verified: the winning hover rule is now the dark chip.

### 4. On-brand content colors regrouped (品牌色收到扩展配色)
`onSolidText` (品牌色上的文字) and `onSolidIcon` (品牌色上的图标) move out of the top accent row into the **Extended colors** section as its first expansion panel (title i18n key `hikari::theme.onSolidGroupTitle`, all 11 locales). The section now renders even when no extension token group is registered. `getDraft()` structure is unchanged — only the editing location moved.

## Verification
- Full suite: **1200/1200 vitest green**, `vue-tsc --noEmit` clean.
- New: `HkImageLightbox.contract.test.ts` (scss source contracts for fixes 1–3 — DOM tests cannot see layout collapse in happy-dom) and `HkMinimap.toolbarOnly.test.tsx`.
- Headless Chrome measurement against the deployed CSS confirms the stage chain fills the frame and the hover cascade resolves to the dark chip.
- Version bump: `0.41.11` (0.41.10 is claimed by fix/root-zoom-popover, v0.42.0 by #455).

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,10 @@
 {
   "name": "@celestia-island/hikari",
+<<<<<<< HEAD
   "version": "0.42.1",
+=======
+  "version": "0.42.2",
+>>>>>>> ee1d16c (🐛 Fix the collapsed lightbox stage, keep its zoom controls and close button visible, and regroup on-brand content colors.)
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,10 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-<<<<<<< HEAD
-  "version": "0.42.1",
-=======
   "version": "0.42.2",
->>>>>>> ee1d16c (🐛 Fix the collapsed lightbox stage, keep its zoom controls and close button visible, and regroup on-brand content colors.)
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkColorSchemeDialog.test.ts
+++ b/packages/vue/src/components/HkColorSchemeDialog.test.ts
@@ -84,22 +84,28 @@ describe("HkColorSchemeDialog extension groups", () => {
   it("renders the extension section reactively when a group is registered after mount", async () => {
     mountDialog();
     await nextTick();
-    // Zero-cost for consumers that register nothing.
-    expect(document.body.querySelector(".s-scheme-groups")).toBeNull();
+    // The section always renders now (the on-brand content group lives
+    // here); before any registration it holds exactly that one panel.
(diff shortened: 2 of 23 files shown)


langyo:
CI queue exemption record (§6.2/§7): self-hosted checks (lint-commits, Test/Format/Clippy) have sat QUEUED beyond the ~15 min window (org runner fleet busy). Local verification is green — vitest 1200/1200, vue-tsc --noEmit clean, plus a green hosted fmt+clippy+test run on the code head. No code-level failures exist; merging per the CI policy.

langyo:
Additional head pushed (ee6e8c5: theme-row lead-cell widening for the chest swatch-overlap report + contract test). CI has not registered check-runs for post-rebase heads (same fleet/mirror lag as above). Local gates on this exact head: vitest 145 files / 1220 tests green, vue-tsc clean. Merging under the recorded exemption.

langyo:
Rebased onto master (bb7392c) after the 0.42.1 publish and retargeted the release to **0.42.2** (0.41.11 would have been a downgrade against the published 0.42.1). Full suite on this head: 148 files / 1279 tests green. Auto-merge remains armed.

PR #460, opened 2026-09-11
🐛 Put the confirm dialog's primary action first and let its window chrome dismiss it.

langyo:
## Reported defects

Two problems in the library's confirmation window, both raised by the product
owner against the emergency-lockdown dialog the 亭 admin panel renders:

1. **The actions row was ordered dismiss-first.** The library now follows the
   desktop message-box convention: primary action first (left), dismiss second
   (right).
2. **The title-bar ✕ did nothing.** `open` is caller-owned, and the component
   only relayed HModal's close request as `update:open`. A consumer that does
   not bind that event (or an awaiting `useConfirm()` caller) therefore saw a
   window that could not be closed, and the promise it was awaiting never
   settled — the operator was stuck with a modal that ignored ✕, overlay
   clicks and Escape.

## Fix

* All three confirmation surfaces now share one order — `HkConfirmDialog`,
  `HkMessageBox` and `HkBlockingToast` (a promise-resolving `alertdialog`, so
  the same category) render the primary action first. Form footers
  (`HkFileBrowserDialog`, `HkModal`'s `footerActions`) keep the primary
  rightmost: they are editing surfaces, not message boxes.
* HModal's close request now runs through the cancel path
  (`onModalUpdate` → `onCancel`), so ✕ / overlay / Escape settle the awaiting
  caller with `false` and take the window down. This is exactly the contract
  `HkMessageBox` already documents for its own dismissal gestures
  ("Backdrop / Esc / title-bar close all count as a cancel"), so the two
  surfaces now behave alike. The `loading` gate still blocks dismissal.

## Tests

New `HkConfirmDialog.test.tsx` (7 cases, the component had none):

* the actions row renders `["Confirm", "Cancel"]` in that order, with the
  primary carrying the configured variant and the dismiss the secondary one;
* the primary action emits `confirm` + `update:open false`;
* the dismiss action emits `cancel` + `update:open false`;
* the title-bar ✕ emits `cancel` + `update:open false` (never `confirm`);
* an overlay click does the same;
* Escape dismisses as a cancel too;
* a `loading` dialog renders no close button, disables both actions and
  ignores the overlay.

`HkBlockingToast`'s behavioural cases now answer by variant rather than by
button index, with its own order assertion updated.

Mutation-checked: reverting the close wiring turns the two dismissal cases red,
reverting the button order turns the three ordering/payload cases red.

## Verification

* `vitest run` — full hikari suite **149 files / 1286 tests passed**
  (one date-picker case flaked once under parallel load and passed in
  isolation and on the re-run; unrelated to this diff).
* `vue-tsc --noEmit` clean; `celestia-devtools verify-versions .` reports no
  drift (cargo=0.3.23, npm=0.42.4).
* npm series bumped to **0.42.4** (0.42.3 is claimed by the open select-panel
  PR); the Rust track is untouched.

## Release note

`0.42.3` is already claimed by two open branches (`feat/select-panel-center`,
`fix/root-zoom-popover`), so this takes `0.42.4` and skips nothing that was ever
published. Publish ordering matters: if `v0.42.4` is tagged before those land, a
later `0.42.3` publish would drag the `latest` dist-tag backwards.

## Downstream

Consumers pick this up through the registry: a chest lockfile pickup follows in
its own PR, and the webui is compiled into that binary, so one rebuild/deploy
brings the new ordering and the working ✕ to the site.


langyo:
### Verification trail (two independent subagent reviews, both SHIP)

**Adversarial round.** Mounted the real `HkConfirmDialog` with hikari's real
`useConfirm()` in chest's exact wiring and drove each gesture:

| gesture | emissions | result |
|---|---|---|
| ✕ | `cancel`, `update:open:false` | `open` false, modal unmounts, awaited promise resolves `false` |
| Escape | same | same |
| overlay click | same | same |

No double settle (✕ + Cancel race, and ✕ followed by unmount, both settle
once — `useConfirm` nulls its resolver). Compensating mutation (restoring the
old passive relay) reproduces the report exactly: `open` stays true, the modal
never unmounts, the promise never settles.

**Ordering/i18n round.** No `:first-child`/`:nth-child` selector, snapshot or
docs prose depends on either row's DOM order; every test selects those rows by
class; all 11 locales carry `hikari::confirmDialog.{confirm,cancel}`;
`verify-versions` reports no drift. It also confirmed there is **no focus
regression** — initial focus is the header ✕ in both orders.

**Both reviews' actionable items are in `2c815ea`:** `HkBlockingToast` (the third
confirmation surface — a promise-resolving `alertdialog`) now shares the order;
the comment no longer claims a library-wide convention; the suite covers Escape
and asserts both actions are disabled while loading.

### Local gates on this HEAD

* vitest — **149 files / 1286 tests passed**; `vue-tsc --noEmit` clean.
* Mutations: reverting the close wiring → 2 red; reverting the order → 3 red;
  the blocking-toast order assertion fails if that row is reverted.
* Commit subjects pass `celestia-devtools commit-msg-lint check`;
  `verify-versions .` reports cargo=0.3.23 / npm=0.42.4 with no drift.

CI is queued behind the congested self-hosted fleet (Clippy/Tests/Run Tests all
queued 7+ minutes; PR Title Check green). The gates above are local for a
Vue-only diff with no Rust surface.

PR #472, opened 2026-09-12
🐛 Read a dragged item's place in the list live, not from the press.

langyo:
## What this fixes

0.43.8 and 0.43.9 taught a drag to **re-anchor on the element** it grabbed, so a list that moves under the gesture keeps the drop on the right item. The **geometry** stayed the press's: the item's line across the strip and its position along it were snapshotted at pointerdown and carried only by the *frame's* movement. Anything that moves the item **within** its frame therefore leaves that snapshot describing a place the item no longer occupies:

* a reorder under the gesture (the element re-anchors, the geometry does not);
* an item added or removed around the held one;
* a font or a display-zoom change;
* a window resize that re-wraps the strip.

The resolution then stops the insertion walk on a phantom midpoint and/or assigns the pointer to the wrong line, so the release **comes to nothing** (or, in the re-wrap case, lands somewhere else). A study measured it at **16.3%** of gestures in reorder scenarios and **17.0%** with no reorder at all — and reproduced it end-to-end in Chromium with the real component: press a chip, cross the 6px threshold, rewrite the host array, release ⇒ `update:modelValue` never fires, with the snapshot **74–311px** off for chips and **90–180px** for panel rows.

## The fix

Read the geometry **off the live strip** instead of carrying it:

`offsetLeft` / `offsetTop` / `offsetWidth` / `offsetHeight` are **layout** numbers — a CSS transform, on the element, on a transition mid-flight, or on any ancestor, does not move them (measured in Chromium, byte-identical). Walking the `offsetParent` chain of the item and of its frame to a common space, subtracting, adding the frame's border back (a positioned `offsetParent` measures from its padding edge) and mapping through the box the frame is **drawn** in (`scale = rect.width / offsetWidth`, plus the frame's own scroll) answers "where is the item I am holding **now**?" — and it is blind to every transform painted on the item, which is exactly what the snapshot existed to avoid.

`offsetLeft`/`offsetTop` are measured from the **padding** edge of an `offsetParent`, which is a fact the frame's border has to be read against: when the frame *is* the item's `offsetParent` the difference of the two offset sums is measured from the frame's padding box, so its border is added back — scaled by the scale the frame is drawn at; when the frame is `position: static` it is nobody's `offsetParent`, the chain skips it, and the same difference is already measured from its **border** box, so adding the border again would count it twice. `layoutOffset` reports whether the walk passed **through** the frame, and the border is added only then.

The carried snapshot stays as the **fallback** for a strip with no layout to read (no layout engine, a detached element, a collapsed box — the guards are per-axis and reject a half-readable or NaN box), so nothing changes where the derivation cannot run.

### Why this and not the other two candidates

A measured head-to-head over **217,998 resolutions** in 8 buckets (clean, our own lift, a host transform on the item, reorder-only, compound, frame/ancestor transform, fractional geometry, give-up shapes), driven through the real composable against the oracle "the item's true current layout rect":

| variant | wrong vs oracle | fixed | broken | gestures that come to nothing | wrong drops |
|---|---|---|---|---|---|
| shipped 0.43.9 | 4,313 (1.98%) | — | — | 4,253 | 60 |
| live re-measure on the re-anchor (+ frame box) | 4,073 | 2,301 | 2,061 | 4,073 | 0 |
| live re-measure **every** resolution | 11,958 | 2,301 | 9,946 | 11,958 | 0 |
| **this revision (transform-free derivation)** | **221 (0.10%)** | **4,092** | **0** | **221** | **0** |

* **Re-measuring the drawn rect is the wrong tool**: it re-exposes the item's own painted transform — the wave-5 class of defect two waves removed. In the compound bucket (host transform on the item **and** a reorder) the shipped snapshot is wrong on 738 resolutions, the re-measure variants on 2,235, this revision on **0**.
* **Dropping the pressed item from the insertion walk** (the other obvious simplification) is exact for single-line strips but changes wrapping semantics — 5.31% drift even with fresh geometry, and it throws the item to the end of the strip when its line holds no other candidate. It also leaves the line (band) staleness untouched, which dominates in wrapping strips. Refuted with 13M swept cases.
* Our own lift can never harm the derivation: `scaleX(1.03)`/`scaleY(1.03)` are centre-anchored on the drag axis, so neither the band nor the midpoint moves (3.6e-15px at every ramp fraction).

## Evidence

* **Gates** (hikari has no eslint gate): `vitest run` **154 files / 1403 tests passed**; `vue-tsc --noEmit` exit 0; `sass` compiles the touched stylesheet.
* **New tests** carry a harness that models what a browser keeps apart — the box an element is *drawn* in (`getBoundingClientRect`) and the box it is *laid out* in (`offsetLeft`/`offsetTop`/`offsetWidth`/`offsetHeight` along an `offsetParent` chain) — from one model, with knobs for the frame's border, its drawn scale, its scroll and whether it is `position: static`. A test can therefore re-lay the strip out, scale it, or paint a transform on it and see which reading the resolution followed:
  * `resolves against where the strip has RE-LAID the held item out` — killed by removing the derivation;
  * `ignores a transform painted on the held item while it follows the layout` — killed by deriving from the drawn rect instead of the offset chain (the wave-5 collapse);
  * `carries the derived geometry through a frame that scrolls` — killed by dropping the scroll term;
  * `reads the frame's border the way the frame hands its layout numbers over` — a positioned and a static frame laid out and drawn identically must answer identically;
  * `resolves exactly as the held item's own drawn box says, across frame shapes` — a differential sweep over nine frames (positioned/static × border/no border × scale 1/1.5 × scrolled × one line/two lines) × every `from` × a pointer grid with probes ON the band edges, compared against `indexAt` handed the item's true drawn layout box. This is what catches the border and scale terms;
  * `keeps the carried snapshot when only half of the box can be read`, `…when the held item collapses`, `…when the layout chain cannot be added up` — each kills a weakened guard, the last one the NaN a chain through an element without offsets (`<svg>`) would otherwise produce.
* **Mutation matrix** — every term is pinned by a test that fails without it: the derivation itself, the offset chain (vs the drawn rect), the frame's scroll, the border (present only through the frame), the frame's scale, the per-axis guards, and the axis mapping.
* **Real-browser validation** (the unit suite cannot cover this: happy-dom reports every box as 0, so it exercises the fallback): a vite dev server + Playwright + real `page.mouse` over the shipped build, re-running the reproduction gestures, the sweeps, a causality mutant and a boundary probe — recorded in the PR thread.

## Contract and residual (deliberate)

* **Fallback**: a strip with no readable layout keeps 0.43.9's behaviour exactly (this is also what the unit suite exercises).
* **Give-ups**: a rotated or skewed frame (this resolution is axis-aligned by construction, like `indexAt` itself); an item lifted out of flow (`position: absolute`); a detached frame or item. Each returns to the carried snapshot rather than guessing.
* **Precision**: the band is the load-bearing value (the line filter tolerates ~0.5px, one-sided); the offset numbers are integers, so the derived band carries ≤~0.6px on chips (0.000px on panel rows in the browser measurements). The midpoint tolerates a full row pitch — which is why a small geometry error on the held item only shows up through the band or in a wrapping strip. A sub-pixel size from `getComputedStyle` is a possible refinement if the band's rounding ever matters.
* The frame's border is added **scaled** by the frame's drawn scale, and the guard that rejects a frame drawn without a box are both exact by construction but only weakly observable: in a single-line strip the held item's own geometry is absorbed by the walk, and a frame drawn zero-wide degenerates the carried snapshot the same way. The border is pinned by the shaped sweep; the frame-box half of the guard is defensive (the component's real frames always draw a box, which the browser run confirms) and is recorded here rather than asserted.
* An independent reviewer measured the **integer rounding** of `offsetLeft`/`offsetTop`/`offsetWidth`/`offsetHeight` against fractional layouts: the derived box can sit up to ~1.3px from the drawn one. A 4.3M-point search for a pointer where that changes a resolved slot found none (the same search finds 192 on the previous revision), and the browser sweeps agree — but it is a property of the primitive, not of this code.
* The derivation must re-read the `offsetParent` chain every resolution: a transform (or `will-change: transform`) on an ancestor **rebinds** `offsetParent`.

## Scope

Only `packages/vue/src/composables/usePointerReorder.ts` (+ its test) changes behaviour; the version goes 0.43.9 → 0.43.10. The composable is not re-exported, so the published surface is unchanged.

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.43.9",
+  "version": "0.43.10",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/composables/usePointerReorder.test.ts
+++ b/packages/vue/src/composables/usePointerReorder.test.ts
@@ -268,6 +268,122 @@ function scrollerStrip(
   return { frame, items, scene };
 }
 
+/** A strip a BROWSER could have laid out: every element reports both of the
+ *  readings a real engine keeps apart — the box it is DRAWN in
+ *  (`getBoundingClientRect`, which every transform moves) and the box it is
+ *  LAID OUT in (`offsetLeft`/`offsetTop`/`offsetWidth`/`offsetHeight` along
+ *  an `offsetParent` chain, which no transform touches). One model drives
+ *  both, exactly as a browser would, so a test can re-lay the strip out (a
+ *  reorder, an insertion, a font change) or paint a transform on it and see
+ *  which reading the resolution followed. */
+function laidStrip(
+  layout: Box[],
+  frameLayout: { left: number; top: number; w: number; h: number },
+  frameDrawn: Box,
+  scene: {
+    scale?: number;
+    scroll?: number;
+    border?: { left: number; top: number };
+    /** The frame is `position: static`, so it is nobody's `offsetParent`:
+     *  the items' layout numbers are then measured from further up — the
+     *  fact that decides whether the frame's border still has to be added. */
+    staticFrame?: boolean;
+  } = {},
+): {
+  frame: HTMLElement;
+  items: HTMLElement[];
+  laid: Box[];
+  relayout: (next: Box[]) => void;
+  setFrameDrawn: (next: Box) => void;
+} {
+  const frame = document.createElement("div");
+  document.body.appendChild(frame);
+  const live = layout.map((box) => ({ ...box }));
+  const pins: Array<() => void> = [];
+  const border = scene.border ?? { left: 0, top: 0 };
+  const sceneOf = () => ({ scale: scene.scale ?? 1, scroll: scene.scroll ?? 0 });
+  frame.style.borderLeftWidth = `${border.left}px`;
+  frame.style.borderTopWidth = `${border.top}px`;
+
+  /** The box a browser draws for a laid-out box inside this frame. */
+  const drawn = (box: Box): Box => {
+    const { scale, scroll } = sceneOf();
+    const left = frameDrawn.left + (box.left - frameLayout.left - scroll) * scale;
+    const top = frameDrawn.top + (box.top - frameLayout.top) * scale;
+    return {
+      left,
+      right: left + (box.right - box.left) * scale,
+      top,
+      bottom: top + (box.bottom - box.top) * scale,
+    };
+  };
+
+  let frameDrawnBox = frameDrawn;
+  Object.defineProperty(frame, "getBoundingClientRect", {
+    configurable: true,
+    value: () => asRect(frameDrawnBox),
+  });
+  Object.defineProperty(frame, "offsetWidth", { configurable: true, get: () => frameLayout.w });
+  Object.defineProperty(frame, "offsetHeight", { configurable: true, get: () => frameLayout.h });
+  Object.defineProperty(frame, "offsetParent", { configurable: true, get: () => null });
+  Object.defineProperty(frame, "offsetLeft", { configurable: true, get: () => frameLayout.left });
+  Object.defineProperty(frame, "offsetTop", { configurable: true, get: () => frameLayout.top });
+  Object.defineProperty(frame, "scrollLeft", { configurable: true, get: () => sceneOf().scroll });
+  Object.defineProperty(frame, "scrollTop", { configurable: true, get: () => 0 });
+
+  const items = live.map((box, i) => {
+    const el = document.createElement("div");
+    frame.appendChild(el);
+    Object.defineProperty(el, "offsetParent", {
+      configurable: true,
+      get: () => (scene.staticFrame ? null : frame),
+    });
+    // A positioned frame is the items' offsetParent, so their offsets are
+    // measured from its PADDING edge; a static one is skipped and they are
+    // measured from the same place as the frame itself.
+    Object.defineProperty(el, "offsetLeft", {
+      configurable: true,
+      get: () =>
+        scene.staticFrame ? live[i].left : live[i].left - frameLayout.left - border.left,
+    });
+    Object.defineProperty(el, "offsetTop", {
+      configurable: true,
+      get: () => (scene.staticFrame ? live[i].top : live[i].top - frameLayout.top - border.top),
+    });
+    Object.defineProperty(el, "offsetWidth", {
+      configurable: true,
+      get: () => live[i].right - live[i].left,
+    });
+    Object.defineProperty(el, "offsetHeight", {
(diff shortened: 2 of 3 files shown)


langyo:
Verification of this revision (frozen hashes), three independent tracks:

**1. Real browser, end-to-end, against the SERVED bytes** (`/mnt/codespace/_scratch/hikari-wave8-verify5/REPORT.md`) — a vite dev server + Playwright + real `page.mouse` over the component:
* the served module's inline-sourcemap `sourcesContent` and its `?raw` text both hash to `e44cd241…`, and removing the five inserted lines from the served original hashes to `6508f06a…` (the previous revision) — i.e. the delivered bytes are exactly those five lines of change;
* the 12 reproduction gestures (press → cross the 6px threshold → host rewrites the array mid-drag → move → release) emit **12/12 with 0 defect no-ops** (0.43.9: 26 across the same sweeps), emitted orders identical to the previous run;
* sweeps: **433 started gestures, 0 defect no-ops, 100% decision and landing agreement** (0.43.9: 12/144 and 9/144 chip, 18/150 row);
* 12 paired control gestures against the released 0.43.9 on identical coordinates: **0 differences**;
* geometry at the release after the rewrite: panel rows `0.000px`, chips sub-pixel (max band 0.609px) against the item's true layout rect; the pre-fix snapshot is 74–311px off;
* the frame's real border path in the live component: the chip frame (`.hk-tag-input-box`, `relative`, 1px border) is in the item's `offsetParent` chain and its border is worth exactly the 1.000px of error it accounts for, scaling to 1.5px when the frame is drawn at 1.5×; the row frame (`.hk-tag-input-list`, `static`, borderless) is not in the chain and needs no term.

**2. Adversarial verification** (`/mnt/codespace/_scratch/hikari-wave8-verify3/REPORT.md`) — real-Chromium re-measurement of the previous round's counterexamples: the border error they measured (−5px at `scale(2)`, +5px static+border, and a resolution flip in a 20px-border static frame) is now **0.0000** and the flipping scene answers 0/20 probes wrong. Three additional findings were fixed in this revision and are now pinned by tests: a chain through an element without offsets (`<svg>`) produced a NaN origin, and two guard halves were untested.

**3. Mutation matrix** — every term is killed by a test that fails without it: the derivation, the offset chain (vs the drawn rect), the frame's scroll, the border (present only when the walk passes through the frame), the frame's scale, the per-axis guards, the NaN guard, the item-collapse guard, and the axis mapping.

Known and recorded rather than hidden: the integer rounding of the offset primitives (≤~1.3px on fractional layouts; a 4.3M-point search found no pointer where it changes a resolved slot, where the same search finds 192 on the previous revision); the frame-drawn-box half of the guard is not observable (a frame drawn zero-wide degenerates the fallback identically) and is defensive only; the guards never fire in the real component (verified with a `Number.isFinite` instrumentation — 17,949 calls, none with a non-finite argument).

PR #475, opened 2026-09-12
🐛 Keep a scaled root's geometry honest in the scrollbar, the timeline and the marquee.

langyo:
## What this does

Two things that belong together: it extracts the **transform-free layout geometry** the drag work had been carrying privately into a shared facility, and it uses that facility on the other places in hikari that were mixing the two coordinate spaces a browser keeps.

### The two spaces

A browser keeps two answers to "where is this element":

* the **drawn** box — `getBoundingClientRect`, `getClientRects`, `Range`, `elementFromPoint` — which every `transform`, zoom, drag lift, FLIP animation and transition in flight moves;
* the **laid-out** box — `offsetLeft`/`offsetTop`/`offsetWidth`/`offsetHeight` walked along the `offsetParent` chain, plus `scrollLeft`/`scrollTop` — in CSS pixels no transform touches.

Comparing one with the other is only correct at scale 1, and this library runs inside consumers that scale the root: chest sets `documentElement.style.zoom` from **100% to 300%**. Measured in Chromium for this change: the four offset values are byte-identical under a transform on the element, mid-transition, and on any ancestor; a transform (or even `will-change: transform`) on an ancestor **rebinds `offsetParent`**; offsets are measured from the `offsetParent`'s **padding** edge, and a `body` `offsetParent` is special-cased by Blink; the four values are **integers** while the computed `width`/`height` are fractional.

### The shared facility — `composables/layoutGeometry.ts`

`frameMetrics` (a box's drawn box, laid-out size and scroll), `layoutOffset(el, frame?)` (the chain walk, with `through` reporting whether it ran through the frame), `laidSize` (the **sub-pixel** border box, content-box padding and border added back), `drawnScale` (how much bigger an element is drawn than laid out — the one factor the older zoom-only `ancestorZoom` cannot see for a `transform` host), `frameScale`, `placePoint` (where a frame draws a laid-out point: its border added only when the walk ran through it, its own scroll taken off in layout units and scaled), `nearestLaidAncestor`, and `layoutRect(el, frame?)`.

`usePointerReorder` — where this machinery was born — now consumes the module instead of its own copy, and reads its two quantities **per half**: the line the item is on and its position along the strip each come from the live layout when the frame can place them, and from the carried snapshot when it cannot. Behaviour where the layout is readable is unchanged (proved by the existing suite and by the browser run in the PR thread).

### The three places that stop mixing the spaces

An audit of all 37 hikari source files that read geometry ranked the mixes by reachability under chest's zoom (k ∈ [1,3]):

| site | as written | what it did at k ≠ 1 | now |
|---|---|---|---|
| `useOverlayScrollbar.ts` thumb drag (attached by 10 components: `HkScrollContainer` by default, `HkDrawer`, `HkTable`, `HkSidebar`, `HkModal`, `HkSelectPanel`, `HkPopover`, `HkMarkdownRenderer`, `HkToolBlock`, `HkErrorLanding`) | `target = dragStartScroll + (delta / trackRange) * maxScroll` with `delta` in drawn px and `trackRange`/`maxScroll` laid out | the thumb travelled **k ×** the pointer: a (k−1)·delta slip, saturating at the track end after 1/k of the drag — past that the scrollbar stops following the pointer at all | the drawn delta is divided by the track's `drawnScale` (k=2: 200 scroll px for a 100-px drag, where it used to run to 400 and stop) |
| `HkTimeline.tsx` natural size (used by 5 chest files) | `if (natural > fitSize + 1) collapsed = true` with `natural` from a probe's `getBoundingClientRect()` and `fitSize` from `clientWidth` | collapsed timelines that fit — a false-collapse band of `(fit+1)(1−1/k)` (200px of a 300px host at k=3) — and, the remembered full size being drawn too, it stayed collapsed | the probe's drawn size is converted into the host's units before the comparison |
| `HkPlaceholderMarquee.tsx` (inside `HkInput` and `HkPasswordInput`) | `--hk-marquee-shift: -copyWidth` from a drawn `copyWidth`, and `copyWidth − 24 > host.clientWidth` | the CSS loop overshot by (k−1) of a copy **per cycle** (a visible backwards jump at every wrap), and the overflow test fired for copies that fit k times over | one copy's drawn width is converted into the host's own units before it is published or compared |

Each fix has a test that fails without it:

| test | mutation it kills |
|---|---|
| `moves the thumb the distance the POINTER went, however the space is scaled` | dropping the space conversion → `expected 400 to be 200` |
| `measures the natural size in the units the host is sized in` | reading the probe's rect raw → `expected 'window' to be 'full'` (the timeline collapses what fits) |
| `converts the copy's drawn width into the host's own units before looping` | publishing the raw rect → `expected '-1000px' to be '-500px'` |

## Evidence

* **Gates** (hikari has no eslint gate): `vitest run` **155 files / 1413 tests passed**; `vue-tsc --noEmit` exit 0.
* **New unit tests for the facility itself** (7): the sub-pixel laid-out border box with and without `box-sizing: border-box` and the integer fallback; the border added only when the walk ran through the frame (both families); the frame's drawn scale applied to what it places; a box that cannot be read (frame drawn with no height, item with no box, a chain through an element without offsets) placing nothing rather than NaN; the drawn-vs-laid scale of an element; and the walk up to the nearest ancestor that has a box.
* **Browser verification** (real headless Chromium + vite dev servers, the shipped package against the same package with only the three fixes reverted — recorded in the PR thread):
  * the zoom model holds: at `documentElement.style.zoom` 1/2/3 a chip's rect reads 72.125 → 144.25 → 216.359 while `offsetWidth` stays 72 and the scroll offsets stay in local px; `drawnScale` reads 1/2/3 and `layoutOffset` is identical at every zoom;
  * a **`transform: scale(2)` host**, which the older zoom-only reader cannot see (computed `zoom` is `"1"`), is read correctly by `drawnScale`/`layoutRect`;
  * the scrollbar thumb: 80 drawn px now scroll 208.5 with the thumb travelling 80.064 drawn (the pointer, exactly); reverted it scrolls 416.5 (2× the pointer) and saturates after 200px;
  * the timeline: `data-mode="full"` where the reverted build collapses to `window`, at zoom 2, zoom 3, a zoomed wrapper and a transform host;
  * the marquee: the published shift is −789.18 (one copy laid out) where the reverted build publishes −1578.36 (drawn), and seeking the animation shows the reverted loop overshooting by exactly one copy width per cycle;
  * RTL (`dir="rtl"`, `scrollLeft` ranging [−800, 0]) and `deviceScaleFactor: 2` behave identically; with no scaling at all the wave-8 drag reproduction still lands 12/12.

## Recorded backlog (audited, not touched here)

The same audit ranked the rest of the class by reachability; these are the ones still mixing the spaces, with the fix shape each needs. They are deliberately **not** in this PR (each changes interaction math in a component with its own behaviour to preserve):

1. `HkImageViewer.tsx` — pan (`panX += dx`) and three anchors (`clientX − rect.left` fed to `setZoom` as a local coordinate): the image travels k × the finger and the pinned point slips by (k−1)·u/zoom. Live in chest (`FileBrowserModal`, `ThemeCustomizeDialog`). Fix shape: divide by `drawnScale(container)` (the container is already the right measuring target).
2. `HkScrollContainer.tsx` `scrollToElement` (`el.rect.top − vp.rect.top + vp.scrollTop`): overshoots by (k−1)·d — at k=3 and d=600 the element scrolls 3600px out of view. The **identical formula is live in a consumer** (chest `ReportsView.tsx` day jump). Fix shape: `layoutOffset(el, vp)` — layout against layout, no division needed.
3. `HkMinimap.tsx` `panDelta` assumes one SVG user unit per drawn pixel (wrong even at k=1 for any card width ≠ the viewBox width, 3× off at k=3); live via chest's SCADA view. Fix shape: the SVG's own `drawnScale`.
4. `HkBoard.tsx` + `useZoomPan.ts` — drawn pointer pixels fed into a board-local camera (anchors, pan, node drag). Latent (exported, no in-workspace consumer yet). Fix shape: `drawnScale` at the viewport element.
5. `HkAuthMethodList.tsx` (a drawn label width published as a local CSS var), `HkWindowedItem.tsx` (drawn rects against a layout overscan buffer — transient, the IntersectionObserver corrects the next frame), `HkPasswordInput.tsx` (a decorative dot canvas sized from a drawn box).

The audit also confirmed what must **not** be touched: pointer-in-one-drawn-box ratios (`HkSlider`, `HkColorPicker`, `HkMediaSlider`, the scrollbar's track click), the layout-only families (`useMeasuredHighlight`, `utils/dom.ts` `pinLeaveGeometry`, `HkTabs`, `HkInput`/`HkTextarea`/`HkAuthCard`/`useSizeMorph`, the scroll-pinning code, `HkMediaVisualizer`), and the drag machinery `HkDraggableList`/`HkDraggableGrid` — whose ghost math is correct under a root zoom (a fixed ghost on `body`, divided by the scale of its own space) and whose slot resolution never reads the dragged element's rect at all.

## Scope

New file `packages/vue/src/composables/layoutGeometry.ts` (+ its test); `usePointerReorder.ts` refactored onto it with per-quantity reads; `useOverlayScrollbar.ts`, `HkTimeline.tsx` and `HkPlaceholderMarquee.tsx` each fixed and tested; version 0.43.12 → 0.43.13. Nothing is re-exported from the package entry, so the published surface is unchanged.

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celestia-island/hikari",
-  "version": "0.43.12",
+  "version": "0.43.13",
   "private": false,
   "type": "module",
   "description": "Hikari Vue 3 component library — production-grade UI components based on shittim-chest design system",
--- a/packages/vue/src/components/HkPlaceholderMarquee.test.ts
+++ b/packages/vue/src/components/HkPlaceholderMarquee.test.ts
@@ -84,6 +84,33 @@ afterEach(() => {
   }
 });
 
+describe("HkPlaceholderMarquee scale-aware measurement", () => {
+  it("converts the copy's drawn width into the host's own units before looping", async () => {
+    // The copy is measured where it is DRAWN while the host it is compared
+    // with (`clientWidth`) is laid out. Inside a scaled or zoomed root the two
+    // differ by the factor the space is drawn at: the raw rect would make the
+    // loop overshoot by (k-1) of a copy per cycle and report overflow for
+    // copies that fit k times over.
+    const { container } = mountMarquee({ text: "a very long placeholder that overflows".repeat(4) });
+    const host = container.querySelector(".hk-placeholder-marquee") as HTMLElement;
+    const copy = container.querySelector(".hk-placeholder-marquee__copy") as HTMLElement;
+    Object.defineProperty(host, "clientWidth", { configurable: true, get: () => 200 });
+    // The host is drawn twice the size it is laid out at: 800 drawn px are
+    // 400 of its own, while the copy really advances 500 of them.
+    Object.defineProperty(host, "getBoundingClientRect", {
+      configurable: true,
+      value: () => ({ width: 400, left: 0, top: 0, right: 400, bottom: 40, height: 40, x: 0, y: 0 }) as DOMRect,
+    });
+    Object.defineProperty(host, "offsetWidth", { configurable: true, get: () => 200 });
+    vi.spyOn(copy, "getBoundingClientRect").mockReturnValue({ width: 1000 } as DOMRect);
+    marqueeExposed(container)?.measure?.();
+    await nextTick();
+
+    const strip = container.querySelector(".hk-placeholder-marquee__strip") as HTMLElement;
+    expect(strip.style.getPropertyValue("--hk-marquee-shift"), "1000 drawn = 500 of the host's own").toBe("-500px");
+  });
+});
+
 describe("HkPlaceholderMarquee", () => {
   it("stays a hidden single-copy probe while the text fits", () => {
     const { container } = mountMarquee({ text: "用户名" });
--- a/packages/vue/src/components/HkPlaceholderMarquee.tsx
+++ b/packages/vue/src/components/HkPlaceholderMarquee.tsx
@@ -1,4 +1,5 @@
 import { defineComponent, ref, computed, onMounted, onBeforeUnmount, watch, type PropType } from "vue";
+import { drawnScale } from "../composables/layoutGeometry";
 
 /**
  * Overflow strategy for a placeholder that does not fit its input.
@@ -64,12 +65,19 @@ export const HkPlaceholderMarquee = defineComponent({
       // copy spacing as trailing padding). Reading the first copy element —
       // instead of stripWidth / 3 — stays correct in the fitting case,
       // where the strip holds a single copy, and through overflow flips,
-      // where the copy count changes under the same strip. The fractional
-      // bounding-rect width keeps the CSS loop shift bit-exact with the
-      // laid-out advance — an integer-rounded width would show a ≤0.5px
-      // seam at every wrap (the copy rides the animated strip, so the
-      // width itself is never affected by the strip's transform).
-      const copyWidth = copy.getBoundingClientRect().width;
+      // where the copy count changes under the same strip. The width is read
+      // where the copy is DRAWN and converted into the host's own units,
+      // because that is the space the CSS loop shift and `clientWidth` live
+      // in: inside a scaled or zoomed root the raw rect is k times the
+      // advance, so the loop would overshoot by (k-1) per cycle and the
+      // overflow test would fire on copies that fit k times over. The
+      // conversion keeps the fractional width that makes the shift
+      // bit-exact — an integer-rounded width would show a ≤0.5px seam at
+      // every wrap (the copy rides the animated strip, so the width itself
+      // is never affected by the strip's transform).
+      const space = drawnScale(host);
+      const factor = space.x > 0 ? space.x : 1;
+      const copyWidth = copy.getBoundingClientRect().width / factor;
       loopWidth.value = copyWidth;
       overflowing.value = copyWidth - COPY_SPACING > host.clientWidth;
       emit("overflowChange", overflowing.value);
--- a/packages/vue/src/components/HkTimeline.test.tsx
+++ b/packages/vue/src/components/HkTimeline.test.tsx
@@ -182,6 +182,45 @@ describe("HkTimeline full mode", () => {
   });
 });
 
+describe("HkTimeline scale-aware collapse", () => {
+  it("measures the natural size in the units the host is sized in", async () => {
+    // The probe is measured where it is DRAWN while the host it is compared
+    // against (`clientWidth`) is laid out: inside a scaled or zoomed root the
+    // probe reports k times the space it really needs, so a timeline that
+    // fits is collapsed — and because the remembered full size is drawn too,
+    // it never comes back. The measurement converts through the space it was
+    // taken in.
+    const proto = Element.prototype as unknown as { getBoundingClientRect: () => DOMRect };
+    const original = proto.getBoundingClientRect;
+    const rect = (w: number, h: number): DOMRect =>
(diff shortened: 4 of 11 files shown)


langyo:
Verification of this revision (frozen hashes, three tracks).

**1. Real browser, end-to-end** (`/mnt/codespace/_scratch/hikari-wave9-verify/REPORT.md`) — two vite dev servers over a byte copy: the shipped package, and the same package with ONLY the three fixed components reverted. Served-module identity confirmed from the dev server (inline-sourcemap `sourcesContent` hashes `75d10428…` and `551da40a…` both match).

* **The zoom model**, measured on a real mounted chip and scroll container at `documentElement.style.zoom` 1/2/3: rect 72.125 → 144.25 → 216.359 while `offsetWidth` stays 72 and `clientWidth`/`scrollTop` stay in local px; `drawnScale` 1/2/3; `layoutOffset` identical {44,66} at every zoom; `placePoint`(90,134) against the live rect (90,133.203).
* **A `transform: scale(2)` host is covered** — which the older zoom-only reader is blind to (computed `zoom` is `"1"`, `ancestorZoom` reads 1): `drawnScale` reads 2 and `layoutRect` is 144.25 wide.
* **Scrollbar** (track 192 laid / thumb 38.39, real `page.mouse`): 80 drawn px scroll **208.5** with the thumb travelling 80.064 drawn — the pointer, exactly; reverted, **416.5** (2× the pointer) and **800 saturated** after 200px.
* **Timeline**: `data-mode="full"` where the reverted build is `"window"`, at zoom 2, zoom 3, a zoomed wrapper and a transform host.
* **Marquee**: shift −789.18 — one copy laid out — where the reverted build publishes −1578.36 (drawn); seeking the animation shows the reverted loop overshooting by exactly one copy width per cycle.
* **RTL** (`dir="rtl"`, `scrollLeft` ranging [−800, 0]): the vertical drag is unchanged, the horizontal drag carries the sign correctly (−306.0 at scale 1, −153.0 fixed at zoom 2 against −306.0 reverted), and the drop-slot reproduction lands (8/8 emissions on the live-geometry prediction). **deviceScaleFactor 2**: every number identical to 1. **No scaling at all**: the wave-8 drag reproduction still starts/emits/lands 12/12, and the three components behave identically with and without the fixes.

**2. Unit tests** — the facility has 7 of its own (sub-pixel border box with and without `border-box` and the integer fallback; the frame's border added only when the walk ran through it, both families; the drawn scale applied to what it places; boxes that cannot be read placing nothing rather than NaN; the drawn-vs-laid scale; the walk to the nearest ancestor with a box), and each of the three fixes has a test that fails without it (mutation-checked: `expected 400 to be 200`, `expected 'window' to be 'full'`, `expected '-1000px' to be '-500px'`).

**3. Audit** — all 37 hikari source files that read geometry were classified formula-by-formula; the eight remaining mixes are recorded in the PR description as a ranked backlog with the fix shape each needs, together with the list that must NOT be touched (pointer-in-one-drawn-box ratios, the layout-only families, and the draggable list/grid, whose ghost math is correct under zoom and whose slot resolution never reads the dragged element's rect).

One pre-existing nit the browser run surfaced, unrelated to this change: the horizontal overlay-scrollbar thumb rests at the track's LEFT end in RTL although the content origin is the right edge — placement is direction-unaware while the drag math is not.

PR #482, opened 2026-09-12
🔥 Retire the standby-pool drill workflow whose backup runners no longer exist.

langyo:
The backup standby runner pool no longer exists: the CI farm moved to node-ci-1..4 and the org runner label set is exactly self-hosted/Linux/X64/local. The drill job runs-on the backup label, so every run is permanently queued and had to be reaped by hand (now auto-cancelled by ci-reaper). A same-fleet health signal already exists: every master push runs real jobs on the same runners.


langyo:
Self-hosted checks are queued on the stalled fleet (workflow-file-only change, no code touched); hosted lint-commits and pr-title-check are green. Merging per AGENTS §6.2/§7 environment-condition exemption.

PR #486, opened 2026-09-12
🎨 Align theme menu host strip rows to the shared lead column.

langyo:
## Problem

Field report (chest, mobile sheet screenshot): the **DPI 控制** entry row in the theme menu is not left-aligned with the theme rows below it — its label sits ~12px further left.

## Root cause

Theme rows widen their leading icon cell to 28px (`.s-theme-item-btn .hk-menu-item-icon.s-theme-item-lead`, #475-class fix) and the customize row was widened separately — both so every row's name starts at the same x. Host rows in the `mode-extra` strip (chest's DPI entry, which correctly reuses `s-theme-item-btn` + `hk-menu-item-icon` + `s-theme-item-name`) kept the generic 16px cell: label x = 12+16+8 = 36px vs the theme rows' 12+28+8 = 48px (sheet: 52 vs 64). The 2026-09-12 scoping deliberately excluded host rows from the widening to avoid glyph rescaling — which fixed the 0.42.2 regression class but left this alignment hole.

## Fix

- Widen the icon **cell** to 28px for rows inside the toggle-owned `.s-theme-mode-extra` strip (no host class-name contract needed), and pin the **glyph** back to `var(--hk-menu-item-icon-box)` — widening a cell must never re-scale its glyph (0.42.2 regression class).
- Contract test: the anti-bare-rule guard is anchored to the rule start (`^…/m`) so the new scoped rule is legal, and a new case pins the strip cell widening + glyph geometry. Mutation-verified (28px→16px turns the new case red).
- Version bump: `@celestia-island/hikari` 0.44.0 → 0.44.1 (CSS-only change).

## Verification

- Independent subagent: geometry check for desktop popover and mobile sheet contexts, glyph-pin specificity/source-order check, mutation red→green, 14/14 tests pass, blast-radius grep clean.
- `vitest run HkThemeToggle.contract.test.ts HkThemeToggle.test.ts` → 14/14 green.

## Follow-ups (audit findings, not in this PR)

Independent audit of chest/hikari row grammar drift recorded in PLAN: PlatformSwitcher add-row bespoke divider/padding, AccountSwitch/LocalePickerModal hand-rolled rows diverging from `mi.item`, host slots (`menu-extra`/`user-menu-extra`/HkMenu header/footer) lacking a row-grammar wrapper, PlatformSwitcher parallel mi.icon/mi.label copy.

---
Rebased onto master (b8c7cd21); head force-pushed to 737bbdf6 to pick up #487. Version rebased to 0.45.1 (npm latest 0.45.0 from #485). Tests re-run green locally (14/14 incl. contract).

### DIFF
--- a/packages/vue/src/components/HkThemeToggle.contract.test.ts
+++ b/packages/vue/src/components/HkThemeToggle.contract.test.ts
@@ -7,12 +7,15 @@
  * assertion: the theme-row lead cell must carry an explicit size that
  * fits the widest host lead mark, not just the mixin's generic box.
  *
- * The widened cell must stay scoped to the lead slot (2026-09-12 field
- * report): shipped first as a bare `.s-theme-item-btn .hk-menu-item-icon`
- * rule, it also matched the 自定义 row and every host row reusing
- * s-theme-item-btn (chest's mode-extra DPI entry) — and mi.icon
- * normalizes those cells' svg to the cell, so their 14px glyphs rendered
- * 28px wide.
+ * The widened cell must stay scoped (2026-09-12 field report): shipped
+ * first as a bare `.s-theme-item-btn .hk-menu-item-icon` rule, it also
+ * matched the 自定义 row and every host row reusing s-theme-item-btn —
+ * and mi.icon normalizes those cells' svg to the cell, so their 14px
+ * glyphs rendered 28px wide. The widening therefore targets named scopes
+ * (lead slot, customize row) plus the toggle-owned mode-extra strip, the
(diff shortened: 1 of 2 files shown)


langyo:
Environment exemption per workspace CI policy: self-hosted queue stalled >10 min (all jobs pending 0s — known farm backlog); hosted fmt+clippy+test green, lint-commits + pr-title-check green, local vitest 14/14 green incl. mutation-verified contract. Merging under §8.3 (CI is reference, not gate).

PR #496, opened 2026-09-13
✨ Add the useImage image loading-state hook.

langyo:
## What

Adds `useImage` — a presentation-free image loading-state composable:

```ts
const img = useImage(() => props.src);
// { src, status: "empty" | "loading" | "loaded" | "error", key, onLoad, onError, retry }
```

Bind `src`/`key` and the two handlers onto an ``, drive placeholder UI from `status`, and call `retry()` after the environment changes (e.g. a resource pack finished installing). Retry re-keys the element so a re-mounted `` issues a fresh request; stale load/error events from superseded elements cannot clobber state.

## Why

Hosts keep re-implementing ad-hoc failed-image handling. Today only `HkImagePreview` tracks load state, and it is a heavyweight preview widget — not a primitive for icons, portraits and avatars. This hook is the missing base layer; a companion `HkImage`-style component can build on it later.

## Provenance

Extracted from langyo/wowsp (PR langyo/wowsp#308), where it is in production across every image site of the desktop app and the marketing site. The API follows hikari composable conventions (string-union status ref, flat return object, `MaybeRefOrGetter` source), which is why the port is verbatim apart from the header comment.

## Tests

Eight vitest cases pin the contract: empty sources, load/error transitions, stale-event guards, retry keying and source-change resets. `pnpm build` (vue-tsc) clean.

Bumps the package to 0.43.0 (new public API).


langyo:
Superseded: force-pushed a rebase onto the current master line (the branch carried a stale 0.43.0 version pin, now dropped — the release flow takes the version from the tag). The pinned base range broke the commit-lint job permanently; reopened as a fresh PR.

PR #499, opened 2026-09-13
✨ Add the runtime registry so every global context and hook self-reports.

langyo:
## The runtime registry — the context of contexts

hikari ships many module-level singletons — contexts (i18n, popup manager, overlay, toasts…), installable document hooks (image fallback, tooltip bridge, safety nets), frame/interval buses and app plugins. Each manages its own state with its own registration API, and until now nothing knew they all existed: debugging "is the image hook actually installed?", "which locale is live?", "is the animation bus alive?" meant knowing each module's internals (user direction 2026-09-14).

### Core — `runtime/registry.ts` (imports only `vue`: no cycles, tree-shakeable)

| API | Purpose |
|---|---|
| `reportHkRuntime(id, { kind, description?, meta?, read?, write? })` | The "I registered successfully" call every subsystem makes on install or first use. Idempotent upsert while active (facets refresh, lifetime/pulse history kept); a fresh lifetime after `dispose()` (epoch guard — stale handles can neither pulse nor dispose the new entry). |
| `handle.pulse(meta?)` / `setMeta` / `dispose` | Liveness beats (`pulses`++, `lastActiveAt`, meta merge); dispose marks the entry disposed and drops its facets but keeps the record listed for post-mortem debugging. |
| `readHkRuntime(id)` | Another subsystem's global state "out of thin air" — no import of the target module. Throwing facets degrade to `undefined`. |
| `writeHkRuntime(id, op)` | Plugin-style steering. Returns `true`/`false`; never throws to the caller. |
| `hkRuntimeSnapshot()` / `listHkRuntime()` / `getHkRuntimeEntry(id)` | JSON-serializable debug dump; stable seq-ordered clones (true registration order). |
| `useHkRegistry()` | Reactive view for debug panels — tracks both the reactive entry states and a structural version, so first-time registrations show up immediately. |

### Auto-reporting integrations — registration APIs unchanged

| id | kind | reports when | read facet | write ops |
|---|---|---|---|---|
| `i18n` | context | module init | locale, key count, merged buckets | `setLocale`, `mergeMessages` |
| `popupManager` | context | first `usePopupManager()` | open count by kind, scroll locks | — |
| `overlay` | context | first `useOverlay()` | open count, names | — |
| `toast` | context | first `useToast()` | slots, messages, types | — |
| `blockingToast` | context | first prompt | pending gates | — |
| `progressDialog` | context | first show/use | open, title, log lines, value/max | — |
| `pageLifecycle` | context | first subscribe/probe | visible, online, listeners | — |
| `imageFallback` | hook | install (dispose on uninstall) | installed, marked count | — |
| `tooltipBridge` | hook | install (dispose on uninstall) | installed, engaged, shown count | — |
| `historySafetyNet` | hook | install | coercion count | — |
| `navigationGuard` | hook | install | blocked count | — |
| `connectionProbe` | hook | first mounted probe | live probes, endpoint | — |
| `animationBus` | bus | first schedule call | entries per tier, paused | `setReducedMotion` |
| `intervalBus` | bus | first schedule | slots | — |
| `cronBus` | bus | first schedule | — | — |
| `errorReporting` | plugin | module init | accepted reports, landing card | — |

Pulses fire per scheduling call / DOM event / state change — never per animation frame. Install-style hooks report on INSTALL, never on import; no-op re-installs keep the first install's truthful meta.

### Consumers

Nothing is required from existing call sites. Hosts may call `hkRuntimeSnapshot()` for support dumps, render `useHkRegistry().entries` in a debug panel, or steer subsystems plugin-style (`writeHkRuntime("animationBus", { type: "setReducedMotion", value: true })`).

### Verification

- 35 new tests (core contract + integrations, incl. cross-context read/write and the safety-net truthfulness guard); full suite **1547 green**, `vue-tsc` clean, shuffle-stable.
- Three independent verification rounds with mutation evidence: pulse counter, popup pulse, structural version — each guard proven "not-fixing-must-go-red", plus a round-3-found coverage gap (toast no-op remove) now guarded.
- Bump: npm 0.46.0 → 0.47.0 (Cargo untouched — npm-only feature, #498 precedent).


langyo:
Merging with the self-hosted checks still queued (farm backlog, environmental per AGENTS §8.3.1): the two required hosted checks (lint-commits, pr-title-check) and the hosted `fmt + clippy + test (hosted)` job are green, and local verification is fully green — 1547/1547 vitest across 5 consecutive full runs, `vue-tsc --noEmit` clean, `--sequence.shuffle src/runtime/` stable across seeds, plus three independent verification rounds with mutation evidence (pulse counter / popup pulse / structural version / toast no-op remove all proven must-go-red). This change touches only the TS package (Cargo untouched).

PR #500, opened 2026-09-13
🎨 Render third-party sign-in methods as independent framed icon tiles.

langyo:
## What

`HkAuthMethodList` (the auth card's third-party sign-in row) no longer renders an `HkIconButtonGroup`: every provider is now its own **framed 44px icon tile** — own border, own `HkTooltip` hover reveal, own focus ring — centered in the card with comfortable independent spacing (`--space-12`).

`HkIconButtonGroup` gets a docs-only update stating its two fundamental jobs: **a selector** (single/multiple) and **a tight action strip** (buttons). npm 0.47.0 → **0.47.1** (rebased onto #499's 0.47.0; 0.46.1 drafted before that release landed).

## Why (user direction, 2026-09-14)

On the chest login page the OAuth row's shared group track measured **446px wide holding two 44px icons** — a full-card-width bordered box with the content floating in the middle. The product owner's ruling:

- the third-party row should be **centered, on demand** — each provider an **independent icon button**, each carrying its own frame;
- **no dedicated button group** for this — the button-group component returns to its more fundamental purposes: (1) a selector, (2) buttons packed closely together. Exactly those two.

So the fix lives in the upstream UI library (this repo), not in the consumer: `HkAuthMethodList` owns the independent-tile grammar, and the group component's contract now says plainly that independent-action rows are not a group. The chest login's MFA factor picker (mode `single`) is exactly a selector and keeps using the group — untouched.

## Contract

Props/emits unchanged (`divider`, `methods[{key,label,icon?,disabled?}]`, `select`), so hosts swap rendering without touching their OAuth flow. Missing icons still fall back to the label initial; disabled entries render dead; every icon-only button carries `aria-label`.

## Verification (3 independent rounds + rebase re-gate)

- R1: 13/13 component tests + **4/4 mutation evidence** (aria-label gutted / disabled removed / constant-key emit / re-introduced `.hk-icon-group` wrapper → each red); full suite green; vue-tsc 0.
- R2: token reality (all consumed custom properties defined in the shipped sheets, every var has an inline fallback), dark-theme behavior, focus-ring grammar vs `.hk-btn`, chest consumer integration (LoginView contract matches; the only `.hk-icon-group` CSS in chest targets the live MFA picker), demo/docs sweep — all clean.
- R3 (final): API surface unchanged, zero CSS collisions, focus order = DOM order with wrap, version hygiene, commit format; bonus mutation (role="group" re-added) → red. MERGE-READY.
- After rebasing onto #499: full suite re-run at the final SHA — **172 files / 1548 tests green**, vue-tsc 0. (One `HkDateTimePicker` chevron test flaked in a single full run, not reproducible standalone on either base or in the re-run — recorded as environment noise.)

A new guard test fails if the component ever regresses into any group markup (`.hk-icon-group` / `role=group` / `role=radiogroup`).

## Consumers

chest picks this up as `@celestia-island/hikari@0.47.1` on its next lockfile refresh (companion PR follows).


langyo:
Self-hosted suite (Run Tests / CI / Clippy…) did not queue for this PR (mirror lag / farm saturation, §8.3.1). Local verification at the merged SHA cb07b2a7: full vitest 172 files / 1548 tests green, vue-tsc --noEmit 0 errors, three independent verification rounds (R1 4/4 mutation evidence, R2 token/consumer audit, R3 final MERGE-READY) — recorded as environment-exempt.

PR #501, opened 2026-09-13
🐛 Keep floating layers sized to their content and clamped to a viewport gutter.

langyo:
## Root cause (user report, 2026-09-14 mobile screenshot)

The "jump to date" FAB tooltip in the chest webui rendered squeezed to ~2 glyphs per line. `tooltipPositionStyle` pinned a `position: fixed` popup whose width was left to shrink-to-fit — i.e. derived from `viewport - left`, the containing block's leftover space. With the trigger near the right screen edge, that leftover is ~2 CJK glyphs wide and the bubble collapses into a vertical column. `width: max-content` was already the fix for the same *class* of bug on the popover panel (#421 ratchet); the tooltip never got it, and neither did the select popout.

## Changes

1. **Content-driven sizing.** `tooltipPositionStyle` now writes `width: max-content` (the `.hk-popover-panel` contract), mirrored in `HkTooltip.scss`. The select popout host gets the same pair (`width: max-content` + viewport cap) — it was the last anchored surface still sizing against `viewport - left`.

2. **Measured clamp for the live popup** — new `runtime/tooltipPosition.applyTooltipPosition()` shared by `HkTooltip` and the document-level title bridge:
   - measure the real box after the base placement,
   - **cap** anything wider/taller than `viewport - 2 × gutter` (max-width reflows the text; max-height + overflow-hidden clips the pathological viewport-tall title — tooltips are pointer-events: none, so clipping beats spilling),
   - **flip** top↔bottom / left↔right when the preferred side is starved (same criterion as HkPopover autoFlip),
   - **shift** the remainder so both edges sit inside the gutter (delta added to the current written coordinates, divided by `ancestorZoom` — chest's root DPI zoom stays correct).
   No-layout environments degrade to the pure placement style (happy-dom pins keep passing).

3. **Shared viewport gutter.** New `--viewport-gutter` L2 token (scale.scss + vendored sync): **16px desktop, 8px mobile** (<768px media query), and `runtime/viewportGutter.ts` reads the same token (px-only, breakpoint fallback) plus `clampWithGutter`. The three private `VIEWPORT_PAD = 8` constants (HkPopover / HkSelectPanel / HkMenu — no desktop variant, tooltip none at all) all collapse onto it, as do the popover panel's CSS caps and the toast container insets. Intentional full-bleed surfaces (drawers, bottom sheets) are untouched — the modal sheet branch keeps `max-width: 100% !important`, which beats the new inline `min()` cap.

4. **Desktop modal frame margin.** `modalFrameMaxWidth()` wraps the resolved width in `min(<width>, calc(100% - 2 * var(--viewport-gutter)))`, so a frame wider than the window can spare no longer runs edge-to-edge.

5. **Versions.** npm series 0.46.0 → **0.47.0**; crates workspace 0.3.23 → **0.3.24** (the `v*` tag namespace serves the npm series per publish.yml; crates publish their declared workspace version).

## Spec mapping (user's three asks)

- *Size by content, cap min/max, overflow handling* → max-content + caps; capped tooltips clip (non-interactive by design); interactive panels keep their existing max-height + overlay scrollbars.
- *Anchor to the document, never sized by the host* → all surfaces teleport to body and size from their own content; the anchor rect only decides position.
- *Strict min viewport margin, 8px mobile / 16px desktop, clamp at edges, drawers/sheets exempt* → the token + unified clamp; HkModalBreadcrumb (centered, always ≥25vw from either edge) and pointer-following drag ghosts are the documented exempt-by-math surfaces.

## Verification (3-round cycle, independent subagents)

- Full suite: **172 files / 1538 tests green** (happy-dom), `vue-tsc --noEmit` exit 0. One unrelated load-flake in the untouched HkDateTimePicker chevron test (green in isolation and on re-run; it hardcodes 2026 and should derive the year — follow-up, not bundled).
- Mutation evidence (red → green, byte-exact restores): M1 remove `max-content` → content-sizing tests red; M2 gutter fallback pinned to 8px → desktop clamp pins red; M3 disable the measurement pass → all 7 layout tests red; M4 remove `applyCaps()` after flip → the two combined cap+flip tests red (a round-1 verifier *proved* the original flip path spilled 33/41px past the edge in exactly those combined cases — fixed and pinned).
- Round-2 adversarial review: mobile sheet `!important` safety, Vue patchStyle clobber safety, opacity-only measurement honesty, popout max-content vs enter-transform/scrollbar-tracks, per-surface spec enumeration (14 surfaces), release mechanics, secrets/debug-leftover scan — all clear; its should-fix (CJK in a comment) and test-honesty nits landed in `d8199eae`.

## Downstream

Chest consumes `^*` registry-first with zero local overrides of `.hk-tooltip-popup` → picks this up on the next lockfile refresh; dev.cw 3005/3009 rebuild rides the usual engine wave.


langyo:
Rebased onto master after #499 (runtime registry) landed: the overlapping `runtime/index.ts` / `tooltipBridge.ts` hunks auto-merged and the combined suite (their registry tests + this branch's floating-gutter tests) is green locally — 172 files / 1573 tests, vue-tsc exit 0. The npm series realigns to **0.48.0** (#499 shipped 0.47.0); crates workspace stays at the branch's 0.3.24 bump.

PR #503, opened 2026-09-14
🐛 Close the half-overwritten history net's option split and derive the datepicker test dates.

langyo:
Leftover sweep (recorded in PLAN): three deferred defects + the CI gap round 2 surfaced, on top of #501.

## 1. History net: half-overwrite option split (recorded defect, PLAN #19 遗留)

`installHistorySafetyNet` claims options only when an install actually patches (the pushState marker is the probe), but the re-patch closure captured the RE-INSTALL call's `fallback`/`evidenceKey`/`options` locals. Scenario: pushState keeps our patch while an external lib restores `replaceState` → a re-install with new options re-patched `replaceState` folding to the NEW fallback while `pushState` and the registry meta kept the first install's — a per-method option split shipping to production with meta describing a different net than the one wired in.

Fix: the closure reads the module-level active capture (claimed only by a patching install), so behavior and meta are structurally incapable of diverging. Regression test reproduces the recorded scenario end-to-end (foreign restore → re-install with a different fallback → the fold target must stay the first install's), including the registry-meta pin via `getHkRuntimeEntry`.

## 2. HkDateTimePicker test: year-rollover time bomb + load flake

The file pinned the literal year 2026 in eight places (deterministic failure on 2027-01-01) and asserted the drilled month view right after a fixed 20 ms settle — which raced the frame/timeout-based drill transition under full-suite load (observed once during #501 verification). Now every expected date derives from one mounted `BASE` constant (local-ISO helpers for the native-input wire format), and the chevron test waits on the drilled view STATE (bounded 2 s poll) instead of a sleep.

## 3. Toast gutter fallback + contract

`inset-inline-end` fallback normalized `1rem` → `16px` (the token's desktop magnitude, matching every other `--viewport-gutter` consumer), pinned by a source-contract test over both toast sheets.

## 4. CI now runs the frontend suite (round-2 finding)

No workflow ever invoked vitest/vue-tsc — the `just test` lane is cargo-only — so all regression coverage lived on the author's machine. New `web-test` job on the existing hosted workflow: frozen install + `vitest run` + `vue-tsc --noEmit` on every PR and master push (public repo; identical install shape the npm-release workflow already runs for `packages/vue`).

## 5. Housekeeping

npm series → 0.48.1, crates workspace → 0.3.25; dropped the duplicate `"version"` key a #501 rebase conflict resolution left in `packages/vue/package.json` (last-key-wins had masked it).

## Verification

- Full suite 175 files / 1577 tests + `vue-tsc --noEmit` green on the final state (independently re-run by round 1 and round 3 verifiers).
- Mutations (independent verifier): closure reverted to invocation locals → the new regression test fails on exactly that pin; toast fallback `16px`→`8px` → contract test fails; chevron expectation un-advanced → chevron test fails at the poll deadline. All restored byte-exact.
- Round 2 adversarial review: no registry-meta consumer in hikari or chest is affected; chest installs with constant options (HMR-safe); publish.yml divergence clause covers tag v0.48.1 ≠ crates 0.3.25; frozen-install shape proven by npm-release.yml.

Deploy notes: npm publish rides the `v0.48.1` tag after merge; chest lockfile pickup follows as usual.


langyo:
The new `web-test` job earned its keep on its very first run: it caught the same drill-transition race class the datepicker de-flake targets, one file over (`HkDatePicker` year-grid test read the title after a fixed 20 ms settle — `expected '2026' to be '2027'` on the CI runner; passed locally every time). `c30919fc` applies the same view-state polling to that test's whole drill chain (year-grid read, title, back-to-days landing).

PR #513, opened 2026-09-14
✨ Add the Tauri-aware credential autocomplete policy.

langyo:
## What

Runtime Tauri-aware autocomplete policy for credential fields — the DOM half of the group-wide "no native credential chrome in Tauri apps" guarantee (the native half is `webview_hardening.rs` in each Tauri app: `ICoreWebView2Settings4.IsPasswordAutosaveEnabled`/`IsGeneralAutofillEnabled` → off).

- `credentialAutocomplete(kind, browserDefault)` composable: Tauri webview → suppression tokens (password `new-password` — the only token the Chromium family respects; `off` is explicitly ignored on login forms — username `off`); browser → component default unchanged.
- `HkSignInCard` / `HkPasswordInput` defaults resolve through the policy; explicit props still win.
- 6 new tests (both modes × three components + prop override); targeted suites green (28+6). The 3 full-suite failures are pre-existing flakes in untouched components (pass in isolation).

## Why both layers

WebView2's fill store is emptied natively (nothing to offer), but WebKit/macOS and plain-browser embeds have no such switch — the DOM contract must stand on its own there.


langyo:
Rebased onto master: the year-grid test fix commit is dropped — superseded by #514's comprehensive settle-race conversion (same flake class, observed deterministically on hosted runners here and fixed locally in parallel).

PR #516, opened 2026-09-14
✨ Standardize pinned scroll content behind the HkScrollPin whitespace contract.

langyo:
## Root cause

A sticky child pinned inside a scroll host whose padding lives on a **wrapper child** loses that padding the moment it pins: CSS sticky resolves its offsets against the scrollport and never sees wrapper padding. HkModal keeps its body padding on `.hk-modal-body-inner` (a child of the scroller `.hk-modal-body-scroll`), so the wizard step header (`HkStepFlow stickyHeader`, consumed by chest `AddProviderWizard`) and the file-browser column head sat flush against the modal header chrome after any scroll — the reported "步骤条钉住了但内边距不见了" bug.

Empirically established in headless Chromium before implementation:

- padding **on the scroller itself** → sticky pins *below* the padding (whitespace preserved automatically);
- padding **on a wrapper child** → sticky pins flush at the scrollport edge (the bug);
- bleed absorption (`padding: <pad>; margin-top: calc(-1 * <pad>)`) keeps the pin's content at its natural position at rest while its own paint covers the padding zone when pinned;
- a full-bleed bottom bar via negative margins keeps working, but must outrank a gutter cover's z-index (see consumer note).

## The contract (new `HkScrollPin`)

**Hosts** (scroll viewports: HkModal body scroller, HkScrollContainer viewport) mark themselves with `.hk-scroll-pin-host` + `data-scroll-axis="vertical|horizontal|both"`, and declare `--hk-scroll-pad-top/right/bottom/left`. Hosts that paint their gutters themselves additionally carry `data-pad-cover`.

**Pins** wrap content in `<HkScrollPin side="top|bottom|left|right" strategy="auto|offset|bleed|none">`:

| strategy | geometry | use for |
|---|---|---|
| `offset` | sticky inset = the declared gutter line; the host's gutter cover masks content sliding under the whitespace | pins anywhere in the flow (toolbars may precede them) |
| `bleed` | absorb the declared padding via padding + negative margin; the pin's own paint covers the padding zone | boundary pins (first/last element of the scroll content) |
| `none` | plain sticky flush | opt-out |
| `auto` (default) | cover host + declared pad → `offset`, else `bleed` | everything |

Dev-only warning keeps **vertical scrollers to top/bottom and horizontal scrollers to left/right** (the axis-side rule from the report). The whitespace travels WITH the pin in every strategy — pinning never eats the padding again.

## HkModal changes

- declares the four pad vars on `.hk-modal-body` (custom properties inherit downward; the body's own `::before`/`::after` **gutter covers** consume them, so pins stop at the gutter line while scrolled content vanishes under the whitespace);
- `.hk-modal-body-inner` padding now consumes the same vars (single source of truth; `--hk-modal-padding-body` still drives everything);
- mobile sheet block retunes the vars to 1rem (covers follow);
- `HkImageLightbox` zeroes the contract for its un-padded immersive body;
- cover paint is overridable via `--hk-scroll-pin-cover-bg`.

## Migrated consumers

- `HkStepFlow stickyHeader` — the timeline root rides the pin contract; strategy resolves at mount (`offset` under a cover host so content above the flow can't be overlapped, `bleed` otherwise); legacy `--hk-stepflow-sticky-top/-z/-bg` knobs keep working, the body gap still folds into the pinned header's padding.
- `HkFileBrowserDialog` list head — pin with `offset`; stops below the body's top whitespace instead of jamming against the window chrome.
- `HkScrollContainer` viewport is now a marked host with its live axis.

⚠️ **Consumer note (downstream repos):** a hand-rolled bleed bar pinned flush at the modal body's bottom edge (chest `HoloDeviceReport` askbar pattern: sticky bottom + negative bottom margin) is now painted over by the bottom gutter cover (z 3) — such bars need `z-index ≥ 4` to keep painting over it. No in-tree hikari consumer is affected.

## Verification

- New `HkScrollPin.test.tsx` (component behavior + dev warnings), `HkScrollPin.scrollcontract.test.ts` (14 source-contract pins, including the full-literal pad-var guard), strategy-resolution behavioral tests in `HkStepFlow.test.tsx`; `HkModal.bodyrhythm/sheetgap` contracts updated to the var-based padding source.
- Gates: `vue-tsc --noEmit` exit 0; `vitest run` 180 files / 1638 tests green.
- Two independent verification rounds (R1 found a guard gap + minors → fixed; R2 mutation-verified all guards RED on break, full suite green, worktree clean).

Version bump 0.49.5 → 0.50.0 (new component).

_Re-checked against master 26233bf1 (#512); local gates re-run green on the merged tree._


langyo:
Superseded by a fresh PR: rebasing onto master de2f8195 (hikari 0.49.7 release line) so the strict up-to-date requirement and the pull_request event delivery both reset. Same single commit, gates green locally (vitest 180/1638 + vue-tsc).

PR #539, opened 2026-09-15
🐛 Keep letter descenders inside every hikari clip box.

langyo:
## What this fixes

`text-overflow: ellipsis` is not a vertical safeguard. The same `overflow: hidden` that truncates a line horizontally clips **glyph ink to the padding box — and that box is the line box**. Where the line box is shorter than the font's ink box, the bottom of every `g/j/p/q/y` is cut off flat. Two user reports of exactly that symptom are already in the tree (the chest header workspace pill from a screenshot, and the `g` tail of an admin-header nickname, patched per-site in #: `HkAdminHeader`), so this PR fixes the library-side sites and adds a guard for the class.

## Sites fixed

| component | before | numbers | after |
|---|---|---|---|
| `HkColorPicker` row-layout label | `line-height: 1` next to the clip at `[data-layout="row"] .hk-color-picker-label` | 12px token → 12px line box vs 14.05px ink (Roboto) / 17.38px (CJK fallback) ⇒ **0.6–0.9px of every descender cut on every font** | `1.5` (18px = the swatch height, so the dense grids do not move) |
| `HkStatCard` value | `line-height: 1.2` + `overflow: hidden` | 24px token → 28.8px box; clears DejaVu/Roboto by 1.06/1.02px, Segoe UI by **0.58px**, and clips **2.9–3.1px per edge** as soon as a host swaps `--font-sans` for a CJK fallback (1.34–1.45em metrics) | `1.35` |
| `HkGaugeRing` centre readout | host utility `leading-none` (`line-height: 1`) | 18px font ⇒ descender **0.95px** below the line box: overlaps the label 2px underneath, and clips inside any truncating host wrapper | new own class `hk-gauge-ring-value` with `1.45` — the ring no longer depends on a utility its host may not ship |

## Drive-by fix: the demo gallery was dead

`just dev` mounted an empty page (2 elements, 0 text nodes): `btoa` throws `InvalidCharacterError` on any code point above U+00FF and the demo's inline SVG contains an em dash ("Hikari — zoom / pan demo"), so `DemoApp` threw during setup. It now encodes through a shared `utf8ToBase64` helper (`TextEncoder` + the existing base64 util). Without this the component gallery cannot be rendered at all — which is also how the clipping class stayed invisible.

## Guard

`packages/vue/src/components/textClip.contract.test.ts` compiles every `components/*.scss` and `styles/*.scss`, resolves the **effective** line-height for each clipping rule *inside its own sheet* (the clip and the line box very often live in two different blocks for the same element — a per-block scan provably misses `HkColorPicker`), and fails when a clip box carries a ratio below `1.3` (the CJK-fallback floor; Latin stacks need 1.10–1.16).

Mutation evidence (the "would it go red" check): reverting `HkColorPicker` to `line-height: 1` makes it fail with
`src/components/HkColorPicker.scss — .hk-color-picker[data-layout=row] .hk-color-picker-label — clips overflow with an effective line-height ratio of 1.00 (< 1.3)`. The shipped tree is green.

## Verification

- `vitest run` — **186 files / 1681 tests pass**, including the new contract test.
- `vue-tsc --noEmit` — exit 0.
- Rendered probe (headless Chromium, glyph ink vs every clipping ancestor's padding box; positive control first — the control page must report 5 findings): the component gallery under `vite dev` reports **0 clipped text nodes and 0 clipped element boxes** before and after the change, i.e. no regression introduced inside the gallery.

## Version

`packages/vue/package.json` 0.54.2 → 0.54.3 (the release job is tag-triggered; tag when you want it published). Panels pick the fix up with their next hikari bump.


langyo:
**Corrections from an independent verification pass** (separate agent, fresh worktrees at `36fd15d8` and at `origin/master`, own harness, mutations). The description above was written from font-metric arithmetic on this box; the rendered measurements below supersede two of its numbers, and three things need recording before this is read as more than it is.

### Numbers corrected
| site | description above said | measured independently |
|---|---|---|
| `HkColorPicker` label | 0.6–0.9px cut | **2.0px** (probe) — 54 ink px lost, 11 of 13 ink rows, ink flat against the box's last row. Roboto is not installed here, so the Roboto figure I computed is not reproducible; the cut is *larger* than stated. |
| `HkStatCard` value | clips only after a CJK font swap | **clips 1.2px with the shipped stack**, no font swap needed (5 ink px, one flat row). The fix's direction is right; the framing understated it. |
| `HkGaugeRing` readout | ink ~0.95px below the 18px line box | **3.0px below** it (with host utilities emulated). Direction right, magnitude understated. |

Confirmed as stated: the `HkColorPicker` fix costs no geometry (row button 26px → 26px, picker pitch 36px → 36px); the guard fails byte-identically on the mutated tree and also fails when copied onto untouched `origin/master` (both findings, ratios 1.00 and 1.20); the demo-gallery fix is real (master `GET /` = 133 elements / **0 text nodes** + `InvalidCharacterError`; branch = 1029 elements / 233 text nodes); gates and commit-message lint green.

### Layout deltas this PR ships (measured, intentional, previously undisclosed)
- `HkStatCard` grows **+3.59px** (95.80 → 99.39px card height; pitch 105.80 → 109.39px) — i.e. every stat grid in shittim-chest (ComputeDashboard, GatewayDashboard, ModelDashboard, DashboardView, CreditsView).
- `HkGaugeRing` inside a `leading-none` host: readout line box 18 → 26.1px, label top 42 → 50.1px; nothing outside the ring moves (wrapper is `absolute inset-0`).
- ⚠️ **Consumer note for the next pickup wave**: chest overrides `.edge-devices__metric .text-lg { font-size: var(--text-xs) }` for its 36px rings (`EdgeDevicesView.scss:134`). With this change that readout's content block goes 36px (exactly 0 slack) → 41.39px (**−5.39px slack**). Chest cannot pick this up until 0.54.3 is published; whoever bumps hikari there should look at that override.
- The readout still carries the host's `text-lg` for its *size* — the ring now owns only the ratio, not the font size.

### Guard limits (so the test is not read as broader than it is)
- It only inspects rules that declare `overflow: hidden|clip` themselves, so **the `HkGaugeRing` fix is not covered by it** (that rule declares no clip).
- The `1.3` floor is looser than the values shipped here (1.5 / 1.35 / 1.45): a regression from 1.35 to 1.30 would still pass.
- The second test (`keeps the review list honest`) is vacuous while the exemption list is empty.

*(The PR body itself could not be edited — `gh pr edit` and the REST PATCH both failed with GitHub-side errors on this repo — so the corrections live here.)*

PR #544, opened 2026-09-16
✨ Let hosts tune the timeline rhythm with gap and label tokens.

langyo:
## Summary

- `--hk-timeline-step-gap` — the node→label breathing room inside a step row (stock 8px).
- `--hk-timeline-label-size` — the step label size (stock `--text-xs`).
- Joins the existing `--hk-timeline-vert-gap` as the third rhythm token; defaults are unchanged, so nothing shifts for current consumers.

Side-column rails (wizards, installers) read better with a notch more air between the number axis and the label and a slightly larger label than compact horizontal wizards; hosts previously had to fight the hardcoded values with descendant selectors.

## Checklist

- [x] Self-reviewed the diff
- [x] Added or updated tests for the changes (existing 18 timeline tests green — defaults unchanged)
- [x] Updated documentation if needed (inline comments at each token)


langyo:
Superseded by #559 (7d279697) — both rhythm tokens (`--hk-timeline-step-gap`, `--hk-timeline-label-size`) are on master since the 0.55.5 line. (Forensics: /mnt/work/hikari-stale-audit/VERDICTS.md)

PR #545, opened 2026-09-16
✨ Let hosts disable individual timeline steps.

langyo:
## Summary

- `TimelineStep` gains `disabled`: the step stays visible so the flow's shape reads, but renders dimmed (45% on node + label), loses `data-clickable` / button role / tabindex, and never emits `select`.
- First consumer: the evernight flasher greys the 铸造凭证 step while signed out — minting needs a session, and now the rail says so instead of routing into a dead form.

## Checklist

- [x] Self-reviewed the diff
- [x] Added or updated tests for the changes (existing 18 timeline tests green; disabled affects only render attributes)
- [x] Updated documentation if needed (interface comment)


langyo:
Superseded by #559 (7d279697) — the `disabled` prop with all render/interaction guards is on master. (Optional follow-up: the dedicated test cases never migrated.)

PR #546, opened 2026-09-16
✨ Give HkModal a headerActions slot before the close button.

langyo:
## Summary

- `HModal` gains a `headerActions` slot rendered at the header's inline-end, immediately before the close button (`.hk-modal-header-actions`: flex row, gap, non-shrinking; the title's existing auto inline-end margin keeps title and actions apart).
- Complements `headerLead` (inline-start) so a header can now carry context on the left and actions on the right.
- First consumer: the evernight flasher parks its connection-settings icon button in the login sheet's header, left of the close glyph, instead of an entry row inside the form body.

## Checklist

- [x] Self-reviewed the diff
- [x] Added or updated tests for the changes (existing 9 modal tests green — additive slot, no behavior change without consumers)
- [x] Updated documentation if needed (inline comments)


langyo:
Superseded by #559 (7d279697), refined by #562 (33088b40) — the slot and its styles are on master with the inline-end pin.

PR #548, opened 2026-09-16
✨ Add HkWaterfall, the shared base view for bucket-grouped card waterfalls.

langyo:
## What

The library-level **base view** for bucket-grouped card waterfalls — the parent the chat waterfalls (reports / history / node list) will share. This PR adds the component only; the consumer migration follows separately so each step stays reviewable.

## Why

The three chat waterfalls each hand-roll the same mechanism, and the duplication is literal: `HistoryView`'s header admits it "mirrors ReportsView's skeleton", and nine blocks (back-to-top thresholds, scroll tracking, day-jump dialog, rAF wiring, empty/loading gates, …) are copied verbatim between two of them while the third re-invents its own. `HkWaterfall` owns the mechanism so those views keep only their content.

## Mechanism

- **Bucketing** — an optional `bucketOf` groups the item stream into ordered sections (the day buckets); without it the list renders as one unnamed section.
- **Columns** — each bucket lays out into `columns` independent flex columns (`1..n`, or `"auto"` from the measured width and `columnMinWidth`). Assignment preserves the reports view's carefully-earned invariant: the stream is walked **oldest → newest** into the currently shortest column and kept newest-first inside it, so a newly prepended (newest) card is the last step of the walk — every older card keeps its column and only the receiving column animates.
- **Windowing** — every card is wrapped in `HWindowedItem` (`estimatedItemHeight`) inside an `HScrollContainer mode="windowed"`, each column carrying its own transition group.
- **Scroll contract** — one rAF-throttled pass publishes the active bucket (last section whose top passed `activeThreshold`) and the opt-in back-to-top signal (hysteresis via `backTopShow`/`backTopHide`); `jumpToBucket(key)` and `backToTop()` are exposed, and the scroll offset is computed against the raw element the tracking already uses so a jump cannot throw.
- **Slots** — `card` (required), `bucketHeader`, `rail` (timeline behind the scroller), `overlay` (floating affordances), `empty`. No content lives in the component.

## Gates

- `vitest run` — **188 files / 1694 tests passed** (8 new: column distribution, the "older cards never change column" invariant, bucketing order, exposed bucket list, jump/back-to-top, empty state, active-bucket + back-to-top from one scroll pass)
- hikari bumps to 0.54.5

The new test caught one real defect while writing it: without `itemKeyOf` the transition group rendered unkeyed children. Every windowed row now gets a key — the stable identity when `itemKeyOf` is supplied, a documented positional fallback otherwise.


langyo:
## Independent verification — round 2 (mutation / "must go red")

Independent verifier, separate throwaway worktree at `03dc5e17`, no installs or codegen runs.

**Baseline first**: 8/8 green (run twice — before and after).

| Mutation | Expected | Actual |
|---|---|---|
| Flip the column walk (newest→oldest + `push`) | red | 1 red (the prepend invariant) |
| Drop the shortest-column scan (always column 0) | red | 2 red |
| Flip the active-bucket threshold (`<= 0` → `>= 0`) | red | 1 red |
| Collapse the back-to-top hysteresis to `top > show` | possibly green | **0 red — coverage gap** |
| Positive control: `buckets` returns `[]` | many red | 7 red |

Verdict: **GO with one NO-GO sub-item**. The verifier proved the hysteresis gap was real (not a harmless mutation) with a throwaway probe that is green on the baseline and red under the mutation.

### Hardening (pushed as `ea94c893`)

- **Hysteresis band pinned** — the new case scripts three scroll positions with `backTopShow: 10` / `backTopHide: 4`: `50 ⇒ [true]`, `6 ⇒ still [true]` and **no new emit** (the band), `2 ⇒ [true, false]`. Self-checked by re-applying the exact mutation: the case goes red with `expected [ true, false ] to deeply equal [ true ]` while the other eight stay green.
- **Column fixture sharpened** — the layout case moves from two columns to three. With two, the mirrored greedy walk produces the same layout, so it could not tell the walk direction apart (the verifier's newest-first variant survived it). Re-checked the same way: under the direction mutation it now fails with `expected [ 'a', 'd' ] to deeply equal [ 'b', 'e' ]`.
- **`jumpToBucket` with an unknown key** pinned as a no-op.
- **Documented** that `backTopShow` gates the whole signal (a lone `backTopHide` is deliberately ignored) — the verifier flagged that silent skip as undocumented.

Still open from the review, deliberately deferred rather than dropped: no cases yet for `columns="auto"` / ResizeObserver measurement, `overscanScreens`, the `rail` / `overlay` / `bucketHeader` slots, or unmount cleanup. The slot coverage belongs with the consumer migration, where those slots get their first real users.

langyo:
## Independent verification — rounds 1 & 3 (correctness + adversarial review)

Two more independent verifiers ran against `03dc5e17` (one of them re-checked mid-flight and saw `ea94c893` land — noted below).

### R1 (correctness) — GO, with one coverage gap

Column distribution was proved **bit-identical to the reports prototype**: the verifier lifted both functions' source, transpiled them, and compared n=0..500 (0 mismatches), plus newest-first ordering and the prepend invariant over multiple column counts. API surface (12 props / 2 emits / 6 exposed) machine-extracted and matched. Its gap: **no case covered "≥2 buckets AND ≥2 columns"**, so a variant that walks columns *across* buckets instead of restarting per section stayed green.

### R3 (adversarial) — BLOCK at the "this is the parent of reports/history" bar; **minimum shipping condition = R1, now fixed**

R3's blocking item was the jump offset, which I fixed in `aaa2d37e` using exactly the primitive it pointed at. It also listed the migration gaps (A2–A7) that are **prerequisites of the migration PR, not of this one** — this PR ships the component only, with zero consumers.

### Fixes in `aaa2d37e`

- **Jump measured in layout space** (R3-R1): `jumpToBucket` now calls the container's `scrollToElement` instead of subtracting drawn rects. chest drives `documentElement.style.zoom` (100–300%), where the old math overshot by the scale factor — about 2× at 200%.
- **Section lookup through `dataset`** (R3-C4): a bucket key containing a quote, backslash or newline made `querySelector` throw, contradicting the component's own "a jump can never throw" wording.
- **`prefers-reduced-motion` honoured for programmatic scrolls** (R3-§5): a JS `behavior` is not reachable by the CSS sheet, so it is read the way `HkAuthCard` reads it.
- **Pending scroll frame cancelled on unmount** (R3-R6): the views this replaces cancelled theirs; the component only stopped listening.
- **Host watched** (R3-R7): a container resolving its viewport after the one-shot mount tick no longer leaves the view silently without scroll tracking.
- **Section keys index-qualified** (R1-§4): a bucket literally named `__waterfall_all__` can no longer collide with the unnamed one.
- **Missing prop docs** (R1-§4): `overscanScreens`, `ariaLabel`, the differing `index` of `bucketOf` (whole stream) vs `itemKeyOf` (inside its column), and the real `activeBucket` contract — an external write syncs the highlight, scrolling then re-derives it, so a mount-time deep link is settled by geometry.
- **New case: multi-bucket × multi-column** (R1 gap) — pins that each section restarts its own column walk. Self-checked by mutation: the whole-bucket-walk variant turns it red (`expected [ [ 'a1' ], [ 'a2' ] ] to deeply equal [ [ 'a2' ], [ 'a1' ] ]`) while the other nine stay green.

Gates after the fix: `vue-tsc` 0 errors, `vitest` **188 files / 1696 tests**.

### Carried forward to the migration PR (from R3, not silently dropped)

Blocking the consumer migration, in its order: ① expose `dockTop`/`dockBottom` slots (HistoryView docks its composer with `--hk-scroll-dock-bottom`); ② a `disabled`/animate-suppression switch (ReportsView suppresses enters while `bulkHydrating`); ③ `estimatedItemHeightOf(item)` or per-card reporting (both views use 2–3 tiers); ④ empty-column collapse / per-bucket column count (2 columns with one card renders a ghost column); ⑤ `data-day-section` compatibility (HistoryView's test asserts it); ⑥ pass through the container's `refresh()` and remaining props; ⑦ bucket spacing and horizontal gutter (the prototype uses 2× and a centred gutter).

PR #542, opened 2026-09-16
✨ Add HkDockBar, the one dock container with edge and corner anchors.

langyo:
## Summary

- Introduces **HkDockBar**, the single dock container every floating surface mounts into: cruise-style bottom-center docks, floating chat bars, corner step pagers, HUD toolbars.
- Chrome is owned by the container (surface, blur, border, radius, reverse glow via `--glow-dock`) so no dock surface ever re-declares its own background again.
- Anchors: `page` (bottom-center floating), `plane` (in-flow inside a host band), and the four corners (`top-left` / `top-right` / `bottom-left` / `bottom-right`) with the shared `--hk-dock-inset`.
- Surfaces: `glass` (default, backdrop blur) and `solid` (moving canvases). Pointer-events pass through the anchor box; the surface catches gestures.
- First consumer: the evernight flasher mounts its wizard step pager (上一步/下一步) on `bottom-right`; shittim-chest's DockBar family collapses into re-exports once this ships.

## Checklist

- [x] Code follows the project's style guidelines (rustfmt / clippy clean)
- [x] Self-reviewed the diff
- [x] Added or updated tests for the changes (9 vitest cases: anchors, hooks, surfaces)
- [x] Updated documentation if needed (component docstring)
- [x] No new warnings introduced

PR #553, opened 2026-09-16
✨ Add HkBlankCanvas, the base view for a full-bleed mount.

langyo:
## What

The base view for views that are **a canvas and nothing else** — the 3D twin, an embedded editor, any surface the host hands to a renderer. W5a of the workspace-panel plan.

## Why it is its own base

The waterfall, kanban, node canvas and data grid all describe *a layout this library draws*. These views describe the opposite: their content is not a layout at all, and modelling it would mean re-declaring whatever the host's renderer already knows. So this component owns only the parts every one of them needs, and draws nothing itself — no background, no placeholder, no error state (a host that needs those renders them in the default slot or the overlay).

## What it owns

- **a 100% × 100% host** — the mount is absolute inside the component, so the renderer sizes itself from the element's box and no padding or scrolling of this component's making can disagree with the surface it draws on;
- **a size-ready gate** — `onMount` is not called until the element has a **non-zero box**. Initialising a WebGL context against a 0 × 0 element is exactly what makes a 3D view flash a wrong first frame and then settle; waiting one frame for layout is cheaper than rebuilding the surface. This is the shape the plan's *"the first frame is already correct"* requirement takes for these views;
- **`onResize` / `onUnmount`** — every later size change (rounded to whole pixels, and not reported when nothing moved), and a teardown signal before the element goes away so a renderer can release GPU resources. `onUnmount` fires only after a successful mount;
- **an overlay slot** — the DOM drawn above the surface (device cards, leader lines, HUD), with pointer events passing through the layer and re-enabled per child so the canvas keeps receiving gestures between the cards.

## Gates

- `vitest run` — **190 files / 1712 tests passed** (6 new: no mount at 0 × 0, mount with the reported size and the `data-ready` stamp, a resize reported as a resize and never as a second mount, no resize when the box did not change, teardown firing only after a mount, and the overlay layer)
- hikari bumps to 0.54.7 — the same version the kanban PR claims; whichever merges second moves to 0.54.8

## Follow-up

W5's second half is the dock: hikari #542 ships `HkDockBar` with four corners plus the two planes, and three of the eight anchors the plan wants (the edge-centred top/left/right) still need adding upstream before the web UI's four hand-mounted dock surfaces can be declared instead.


langyo:
## Independent verification — GO

Correctness checked on the tree: the size gate blocks on **either** side being zero (probed three ways: (0,600), (800,0), (1,1)); `data-ready` is set inside the same frame as `onMount` but lands in the DOM one microtask later; a later 0 × 0 is reported as a resize rather than a second mount; `onUnmount` fires only after a mount; the overlay sits above the mount with the source-level pointer contract intact; the root contributes no padding or scrolling. Regression: `vue-tsc` 0 errors (with a deliberate-error positive control) and **190 files / 1712 tests green**, against an independently reproduced master baseline of 189 / 1706.

Mutations M1–M3 and M5 turned exactly the expected cases red.

### Follow-up recorded, not silently dropped

The review's one substantive gap: **the stylesheet-level contracts have no case** — removing the overlay's `pointer-events: none`, or its child rule, or reversing the layer order all stay green. The repo already has four `sass.compile` contract tests (one of them asserting exactly this property), so this is a deviation from house convention rather than an untestable area. Also unpinned: the gate's "either side" semantics, the later-0 × 0 behaviour, `ro.disconnect()`, `ariaLabel`, and the fact that `renders the overlay above the mount` asserts only the slot's presence, not the layer order.

A contract test covering those belongs in a small follow-up rather than expanding this PR.

PR #554, opened 2026-09-16
✨ Add HkNodeCanvas, the camera the node surfaces share.

langyo:
## What

W4a of the workspace-panel plan: the base view for a pannable, zoomable node surface — and the last of the four base views (waterfall #548, kanban #552, blank canvas #553).

## The abstraction is the camera, not the drawing

The industrial topology, the media pipeline and the topic map are the same shape: a camera over a piece of content, drawn as SVG, as DOM over an SVG edge layer, or as DOM over a canvas. What they share is **not how they draw** — it is the camera:

- how a wheel tick becomes a **grid-snapped** zoom step that lands on a reproducible value;
- how a drag **pans**, and how the point under the cursor stays fixed while zooming;
- how **fit** frames the content (with padding, a lower clamp and a cap so a small graph is not blown up);
- how a **screen point maps back into content coordinates**.

That rule currently lives in three places (the scada camera module, the pipeline's own viewport, the topology layout). This component owns it once.

What stays out: **edges, nodes, routing and layout** — the host draws those through the default slot, which lives *inside* the transformed layer, so the host thinks purely in content coordinates. Overlay chrome (inspectors, rails, HUD) lives outside it and gets screen coordinates.

## The first frame is already correct

This is the requirement stated as code: when `fitOnLoad` is on and the content bounds are known, the camera is computed **before the first paint** rather than starting at 1:1 and snapping a frame later — the flash a large graph used to open with. Bounds that arrive after mount (the usual case: the graph loads later) are framed the same way, still before the host has drawn at the wrong scale.

## The minimap is on by default

Bottom-right, the corner the chat views use, with the placement selectable. The component owns **where** the map sits; the host owns **what it shows** — the `minimap` slot receives the live camera, so a host feeds its own boxes without reaching into the component.

## Independent verification — round 1 was a NO-GO, fixed here

The first review of `62788ea4` came back NO-GO with four defects, each one reproduced by a failing assertion before it was fixed:

- **D1 — the wheel was dead between k = 0.20 and k = 0.45 (P1).** A `zoomFactor` of 1.05 is smaller than half a grid step below k = 0.5, so multiply-then-snap rounded straight back to where it started: twelve consecutive wheel ticks moved the camera by exactly nothing. The band is not exotic — any content wider than ~1600 world units is fitted into it (the 4000-unit fixture the first test pins sits at k = 0.20).
- **D2 — capturing the pointer on pointerdown killed card clicks (P1).** Capturing retargets the gesture's click to the capturing element, so clicks on node cards never reached their handlers. This workspace has already paid for this once (shittim-chest #818, the timeline gesture); capture now waits until the pointer has actually travelled 6 px, so a tap stays a tap.
- **D3 — fit was computed against a scale it did not render (P2).** `x`/`y` came from the un-quantised fit factor while the camera rendered the snapped one, so 100×2300 content overflowed the bottom edge by 15 px and a centred graph sat 40/60 instead of 50/50. Fit now snaps **down** first — rounding up can only ever exceed the padding budget — and derives the translation from the scale it actually renders, which also keeps binary dust out of the camera the host shares (`6 * 0.05` is `0.30000000000000004`).
- **D4 — late bounds threw the user's camera away, with no way out (P2).** The bounds watcher re-fitted unconditionally, so panning before the graph arrived silently reset the view — and this is the *documented* usual case, not an edge case. There was no opt-out either: the only switch, `fitOnLoad: false`, disables the first-frame fit as well. Automatic framing now stops as soon as the camera has been moved by hand.
- **D5/D6 (P3).** A canvas mounted inside a hidden tab is framed when it becomes measurable; degenerate bounds are a real no-op instead of rewriting a host-controlled camera.

The same review found the coverage behind those defects: all nine original fixtures had `x = y = 0`, so the two centring terms could be deleted with the suite still green, and the test named *"does not zoom when zooming is off"* dispatched no event at all — deleting the `zoomable` gate outright kept it green.

## Independent verification — round 2 was a NO-GO too, fixed here

The re-review of the fixes above found that two of them had introduced fresh regressions, plus three P1 contract problems:

- **`fit` threw away a whole grid step (D7).** `720 / 1200` is exactly `0.6`, but in binary it is `11.999999999999998` grid steps, so the new floor landed on `0.55`: a 1200×800 graph was framed 8% smaller than the padding budget allows, with 70 px of side padding instead of 40. The floor now carries an epsilon.
- **A released pan kept following the pointer (D8).** Deferring the capture to the 6 px slop meant the element never saw a `pointerup` that happened outside it, so a press near the edge followed by a release outside left the pan armed — and the next button-less hover moved the camera 200 px on its own. `onPointerMove` now drops a gesture whose `buttons` is 0.
- **A host that spells `contentBounds` as an object literal looped (P1).** `v-model:camera` plus a per-render bounds object meant fit → emit → host write-back → new identity → watcher → fit, until Vue threw *Maximum recursive updates exceeded*. `fit()` now compares cameras **by value**, so a re-fit with nothing to change is silent.
- **Automatic framing fought a host-supplied camera (P1).** `fitOnLoad` overwrote the `camera` prop at mount and on every bounds change, although the prop's own documentation says it seeds only the component's own camera. Every automatic path now goes through one `autoFit()` that skips a controlled camera entirely; `fit()` stays the host's imperative re-frame.
- **The root's capture cut host node-drags short (P2).** Node dragging is the core gesture of all three consumers, and root capture retargets every subsequent pointer event. A host can now claim the gesture by calling `preventDefault()` on its own card's `pointerdown`.
- **`touch-action: none` was unconditional (P2)** — it moved to an inline style that is `none` only when `pannable || zoomable`, so a surface that consumes nothing gives the page its scrolling back; and a fit that cannot be described (degenerate range, `minZoom: 0`) now leaves the camera alone instead of putting `NaN` into every derived coordinate.

Still open, deliberately, and recorded here rather than fixed in a hurry: `movedByHand` is one-way and is also set by host imperative calls, so a host zoom toolbar can disable automatic framing for good; there is no pinch-zoom and no keyboard path; a *controlled* camera can carry an out-of-range or non-finite `k`; and a controlled host must write the emitted camera back in the same tick.

## Independent verification — rounds 5 and 6: two more NO-GOs, and where this now stands

Rounds 5 and 6 (the latter driving the real component in headless Chrome 153 over CDP, with red/green pairs against the previous commit) confirmed every fix in this PR and then found that the *controlled-camera* contract has one remaining hard case.

**Confirmed fixed in real Chrome:** the fit floor (zero-floor uses the un-quantised ideal, four-edge padding holds, an independent oracle agrees over 72 combinations), the echo rule (rounding / clamping / coarse-snapping / 10 Hz-mirror / object-replacing stores all converge — a jittering store went from 1501 emits to 1 — and each of them is answered after it moves its own camera), the gesture base (a host's restored `scale(2)` view survives a drag), the element-blur regression (pressing the canvas while a host field has focus no longer kills the pan; a real window blur still ends it), the `Infinity` guards on both the zoom and the translation, the chrome guards and control handback, and listener hygiene (one capture pair per live pan, nothing leaked across unmount).

**Round 10's finding, fixed in `26733ac6` — a host whose write-back is asynchronous and in order.** With a store that echoes each write ~100 ms late carrying an *older* value, a 300 px drag moves the view 100 px and the emit chain visibly rewinds; the parent commit moved it the full 300 px monotonically. The arithmetic is that `gestureBase()` falls back to the echo, and that echo is behind the pointer. The obvious one-line revert was measured and **is not sufficient**: dropping the outstanding-request condition fixes this case but re-opens the pan-limit dead zone and the stale-base teleport. The base now carries an explicit list of the requests the host has not answered, so "still catching up" can be told from "answered" — see the round 10 section below.

**Round 10's finding, fixed in `26733ac6` — a host that never applies a write.** `fit()` returns `false` for ever, and `false` cannot be told apart from "nothing to change". This had been carried since round 3 as the one finding no fix in this PR addressed; an unanswered request now expires, so a later `fit()` asks again.

**Test honesty.** The 44 tests pin only the blur change: run against the previous source, 43 of them still pass. A comment claims browser-probe coverage for the deleted window-blur test, and no such probe exists in this repository. The five tests added in `26733ac6` pin them instead: run against `c95b155b` they catch four of the last changes plus the blur guard, and run against `7ad93728` they catch the in-order regression and X03.

## Round 10: the in-order store and the host that never applies — fixed in `26733ac6`

Round 10's two survivors are fixed in `26733ac6`, which replaces the boolean "a request is outstanding" with an explicit list of the requests a controlled host has not answered yet.

**The gesture base is the newest request the host has not answered.** Every write to a controlled host records `{ camera, framing, at }`, and a host write is matched against that list:

- it reproduces an **older** in-flight request: a store behind a queue is still catching up, so the newest request stands and the gesture keeps composing from it. This is the round 10 regression — a 300 px drag against a store that echoes every write ~100 ms late, in order, now lands **300 px monotonically** (head: 100 px with visible rewinds; the parent commit also landed 300 px, which this keeps).
- it answers the **newest** request — matched exactly, or rounded, clamped, snapped, interpolated — so the host has spoken and `camera.value` is the base again. That is what keeps the pan-limit case working: 200 px into a wall and 200 px back moves the view **−200 px at once** (head and parent agree; the one-line revert measured in round 10 re-opens the dead zone, this does not).
- nothing is outstanding: the host moved itself (a "reset view"), so the next `fit()` answers.

A **`fit()` request is not a gesture.** It is absolute and one-off, so it is a base only while it is still plausibly in flight (a 250 ms grace) *and* only for a host that has shown it applies what it is asked to. That is what stops a dropped framing request from teleporting the surface: round 9's `dropFirst` + 80 px drag case went from **465 px at `scale(0.3)`** (head *and* the parent commit) to **80 px at `scale(1)`**, and one wheel tick right after the same dropped fit now zooms about the camera that is actually on screen (`{k:1.05, x:-20, y:-15}`) instead of one the host never had (`{k:0.35, x:382.5, y:25.075}` at head, `y` jumping 525 px at the parent).

**X03 is fixed, not documented away.** A repeated identical `fit()` no longer stands for ever: while the host has not answered, the request expires after the same 250 ms grace, so a later explicit `fit()` asks again and returns `true`. `false` now means exactly "no write was made" — the camera is already the fitted one, the same request is still inside its grace, or the host answered it. An *answered* request still stands, which is what keeps a host that frames on every update from looping when the store rounds, clamps or jitters (one request, one emit).

Real Chrome 153, a `dropFirst` host with `fitControlled` over 100×1571 bounds in 800×600 — head: `fit()` at +120 ms `false`, +270 ms `false`, +720 ms `false`, one emit, camera `{1,0,0}`, never framed. Fixed: `false`, **`true`**, `false` (nothing to change by then), two emits, camera **`{0.3, 385, 64.35}` — framed**. A `dropAll` host goes `false, true, true` and keeps asking; the camera stays where that host puts it, which is its own contract.

Residuals, stated rather than implied: inside the grace a repeat is still refused (the request may yet be applied — that is the point of the window), and a host write that arrives while requests are outstanding is still read as the answer to the newest one (round 10's NEW-2), so a host that moves its own camera in that window needs one more write before an identical `fit()` is answered.

**The tests pin the changes instead of describing them.** Five tests are added, and each goes red when its behaviour is reverted:

| behaviour | test | red at |
|---|---|---|
| gesture base across a lagging in-order store | `keeps the drag on the newest request when the store echoes older ones` | `7ad93728` (2 rewinds in the emit chain), and on reverting the base rule |
| the echo rule, answer side (pan limit) | `re-anchors the gesture to the camera the host actually kept` | `c95b155b`, and on reverting the re-anchor |
| the echo rule vs a transforming store | `re-frames after a rounding store restores the camera it had before the last fit` | `c95b155b` |
| the finite-translation refusal | `refuses a fit whose translation would overflow` | `c95b155b`, and on deleting the guard |
| X03 | `asks again when the host never applied the request` | `7ad93728`, and on restoring the for-ever guard |

Run against `c95b155b` the new file has 5 red where the old 44-test file had 1 (the element-blur guard); run against `7ad93728` it has the in-order regression and X03.

## Gates

- `vitest run` — **199 files / 1884 tests passed**, 49 of them this component's (9 before the fix rounds; they now cover centring on non-origin bounds, the padding budget for off-grid bounds, a monotone wheel in both directions down to `minZoom`, tap-versus-drag capture, a release that happens off the canvas, a host that claims its own pointerdown, late bounds with and without a hand-moved camera, the hidden-tab reveal in both states, the degenerate-bounds no-op, and a host write-back loop)
- hikari bumps to **0.54.11** — #552 and #549 took 0.54.9 and 0.54.10 while this was in review

## Follow-up

Consumers (topology, media pipeline, topic map) migrate next; their per-view specifics — edge routing, node sizing, layer filters — are what the default slot carries.


langyo:
**Round 11 — round 10's two survivors are fixed in `26733ac6`** (PR description updated with the rule, the measurements and the residuals).

* **NEW-1 (gesture base).** The boolean "a request is outstanding" is now an explicit newest-first list of the requests the host has not answered. A host write that reproduces an *older* in-flight request is a queue still catching up, so the newest request stands and the gesture composes from it; a write that answers the *newest* request (matched, rounded, clamped, snapped — anything) re-anchors the base to the host's camera. A `fit()` request is excluded from that chain: it is absolute and one-off, so it is a base only while it is plausibly in flight (250 ms) and only for a host that has shown it applies our requests. Real Chrome 153, `queue100` (100 ms in-order echo), 300 px drag: **300 px, monotone, 0 rewinds** (head `100 px` + rewinds; parent commit 300 px). Dead zone stays fixed (back-drag after the wall: **−200 px**), and the dropped-fit teleport round 9 carried is fixed too (`dropFirst` + 80 px drag: **80 px at `scale(1)`**, was 465 px at `scale(0.3)` at head *and* at `c95b155b`).
* **X03 (fixed, not documented):** an unanswered identical request now expires after the same 250 ms grace, so a later explicit `fit()` asks again. Real Chrome, `dropFirst` host: head `fit()` at +120/+270/+720 ms = `false,false,false`, one emit, never framed; fixed = `false,true,false` (framed by then), camera `{0.3, 385, 64.35}`. `false` now means "no write was made": already framed, still inside the grace, or the host answered. Answered requests still stand, which is what keeps a fit-on-every-update host from looping with a transforming store.
* **Tests:** five added, each red when its behaviour is reverted (mutation-checked one at a time, source restored byte-for-byte). Run against `c95b155b` the file has 5 red where the old one had 1 (the element-blur guard); against `7ad93728` it catches the in-order regression and X03.
* **Residuals, stated:** a repeat inside the grace is still refused; a host write that lands while requests are outstanding is still read as the answer to the newest one (round 10's NEW-2), so a host that moves its own camera in that window needs one more write before an identical `fit()` is answered.
* **Gates at `26733ac6`:** `vitest run` 199 files / 1884 tests passed; `vue-tsc --noEmit` 0 errors.

Driving it in headless Chrome needs the app from `round 10`'s scratch tree; the two shapes worth re-running are `e7` (`lag100`/`queue100`) and `e3f` (the wall), plus `e3b` (the dropped fit).

PR #555, opened 2026-09-16
🔗 Align the i18n documentation with the locale codes the code actually uses.

langyo:
Locale-naming unification, docs-and-comments only (12 files, 26/26 lines).

`packages/i18n` emits and consumes **`zh-Hans`/`zh-Hant`** everywhere — `Language::code()`, the embedded locales directory names — but the i18n doc tables in eight languages, the migration guide's translation-target list, a stale dynamic-loading example (`ZH_CHS_TOML`/`ZH_CHT_TOML`), and the `ThemeProvider` doc comment + example still said `zh-CHS`/`zh-CHT`: codes the library has never emitted. This aligns all of them with the code.

No code path changed (the two `provider.rs` hits are a doc comment and a `no_run` example string). Part of the workspace-wide Chinese-locale unification.


langyo:
Merging with the required check (`lint-commits`, strict) green. The Clippy/Test/Format/vitest jobs are non-required and queued behind a saturated farm; this PR is Markdown tables and a doc comment plus a `no_run` example string — no code path changed — so they gate nothing here.

PR #556, opened 2026-09-16
🐛 Steady the modal fold under streaming content.

langyo:
Closes the two remaining window-motion defects from the 2026-09-16 chest report: windows sometimes entering "from below" instead of the center unfold, and the close destroying the content before the fold.

Frame-level probes against the deployed demo (viewport 1440×900, the todo-log window over the cruise stream) pinned both mechanisms before the fix:
- close: the frame collapsed 697px/1154 chars → 200px/24 chars in the SAME tick as the close (consumer teardown), and the fold then played over the emptied stub;
- enter: content streaming in mid-enter snapped the frame 459→697px in one 3ms step — the desktop choreography's geometry is height-relative (translateY 5% + bottom-10% clip of the frame height), so the unfold was recomputed under the running animation; under frame starvation the CSS transition latched late and the machine's flip-relative deadline cut it mid-flight (enter frozen at opacity 0 for ~640ms, then a snap).

### Close fold: hold the last live content (`useSurfaceContentHold`)
Consumers routinely tear down the state feeding a window's body in the same tick as the close. While the machine is in a closing phase the surface now serves the LAST live-rendered children (Vue's patch skips identical vnode references, so the held DOM is never re-patched mid-fold). Wired into HkModal (header+body+footer) and HkDrawer. A reopen interrupt drops the hold with the phase; `closed` drops the cache.

### Enter unfold: freeze the frame height (`useSizeMorph` `deferRemeasure` + early arm)
HkModal arms the size morph WITH the enter (one tick after the mounting patch) and gates resize-driven re-measurement through the opening phases; the open edge flushes the deferred growth with an explicit remeasure one tick after the transition classes clear, so it animates instead of snapping. Post-fix probe: the enter plays on a frame whose height never moves during the choreography, and the growth lands as a smooth post-enter morph.

### Leave fold: keep the pin (`SizeMorph.hold()`)
The close choreography owns a stable box for the whole leave: `hold()` disarms the observer/timers and cancels any in-flight reveal but KEEPS the height pin (probe: leave heights frozen at a single value every round), released on the finalize edge; `stop()` after a hold still releases, and a never-armed stop stays the no-op it always was.

### Machine: re-arm the deadline on `transitionstart`
A `transitionstart` on a layer element (target-filtered, one re-arm per element per phase) re-arms the deadline from the actual latch moment, so a late-started (frame-starved) transition runs to completion instead of being cut mid-flight. A never-started transition fires nothing — the flip-relative deadline stays the bound (A2 unchanged).

## Verification (3-round independent loop, all SHIP)
- vue-tsc clean; full suite **193 files / 1725 tests** green on the rebased base (master's 0.54.8 included).
- Six mutations each turn their test red: hold-neutralized, armStart no-op, hold-releases, defer-gate-ignored, hold→stop wiring, open-flush removed.
- Frame-level browser probes (local build bundling this branch, backend proxied from the demo face): close starts with the open-state content in 6/6+4/4 rounds with a single frozen leave height; enter height constant through the choreography with the growth smoothed post-enter; starved rounds settle at the latch-relative deadline instead of the flip-relative one.

## Notes
- Known accepted micro-race: closing within the ~150ms post-enter growth animation lets the running height transition finish under the fold (the pin is already at target); content and fold stay intact.
- Desktop modals have no bottom-docked code path: the sheet family is media-gated (≤767px) and the only JS `innerWidth<768` gate is geometry-neutral — the perceived "from below" was the cut-mid-rise snap plus the mid-enter height jump, both fixed here.

packages/vue 0.54.8 → 0.54.9.

---
*Rebased onto 0.55.0 (10803d37): version conflict resolved to 0.55.1; full suite re-run green on the new base (198 files / 1835 tests, vue-tsc clean).*


langyo:
CI 环境性豁免（AGENTS §8.3.1）：自托管作业排队 >25 分钟无人认领——农场现状 node-ci-2 offline、node-ci-1/3/4 均忙于其他 org 作业（ci-farm-watch 台账 19:40）。本 PR 为纯 TS 变更（packages/vue），代码级检查 `web: vitest + vue-tsc (hosted)` 已 PASS；本地门禁 vue-tsc 0 错 + 全量 193 文件 / 1725 测试全绿（rebase 后基线 0.54.8 上复跑）；lint-commits 与 pr-title-check 均 PASS。排队中的 Rust 作业（clippy/test/fmt/coverage）不触及本变更的任何文件。

PR #563, opened 2026-09-16
✨ Render the password hold-to-reveal as counter-drifting noise that screenshots cannot read.

langyo:
## What

Replaces the anti-OCR hold-to-reveal of `HkInput variant="password"` (`passwordTrailing="eye"`). The old pass drew the password as text with per-glyph jitter — every single frame was still legible, so a screenshot captured readable text. The new pass is a **random-dot kinematogram**:

- one shared noise tile fills the whole reveal area;
- the noise sampled **through the password glyphs** drifts one way, the **background** noise drifts the **opposite** way (±40 CSS px/s);
- a human reads the glyphs off motion coherence; **any single frame — a screenshot — is statistically pure noise**: glyph zone vs background zone measured mean 170.6 vs 171.7, std 54.3 vs 54.6 (device px, real Chrome).

The glyph mask is rasterized only on an **offscreen** canvas and immediately re-composited with noise (`source-in`); the visible canvas receives nothing but two noise pattern fills and one `drawImage` of the composited mask — the invariant the new per-canvas recorder test pins (`fillText` with password content may never target the visible canvas).

## Evidence (real Chrome, headless, dpr 2)

Repro: probe page + driver archived at `/mnt/work/reveal-probe/` (page/probe-reveal.html, page/probe-reveal.ts, run.mjs; evidence: measure.json, held-frame-*.png, reconstruction.png; independently re-run in R1 → `/mnt/work/reveal-probe-r1/` and R2 → `/mnt/work/reveal-probe-r2/`).

- **Single-frame safety** — glyph band vs background band: R1 mean Δ 0.43 / std Δ 0.45; R2 mean Δ 0.166 / std Δ 0.003. No glyph structure exists in any one frame.
- **Coherent background drift** — two-frame cross-correlation peaks at pixel-perfect **1.000** (R1: s=+22 px/251 ms; R2: s=+21 px).
- **Glyph counter-drift (the human-readable signal)** — averaging 24 frames aligned to the *background* drift makes the glyph mask emerge as structure (integration std ratio bgAligned/unaligned: R1 **1.605**, R2 **2.010**, vs glyph-aligned ≈ unaligned). That temporal integration is exactly how motion segmentation reads the field, so what the reconstruction shows is what a human sees. `reconstruction.png` carries the visual proof (R2 decoded it: structure ratio 2.53, glyphs emerge only in the bg-aligned panel).
- Video correlation can still recover the drift fields in principle — motion must exist for the human, so it exists for software; that residual risk is the accepted trade-off for a user-initiated hold-to-reveal.

## Degradation paths

- **Reduced motion** (animation bus parked): the new `isAnimationParked()` getter makes the degrade **immediate** — no 160 ms of unreadable noise first — falling back to the legacy **static** per-glyph jitter drawing. Flipping reduced motion on **mid-hold** is covered too: a recurring cronBus watchdog re-checks while held.
- Frames never arriving although the bus is not parked (hidden document, extreme jank): the same recurring watchdog (bare timers, fires when rAF cannot) degrades within ~160 ms, latched for the rest of the hold. Known safe-direction edges: a hold started mid-scroll can latch to the fallback early; background tabs throttle the check to ~1 s. Both degrade to the previous readable behavior and self-heal next hold.
- Engines without canvas patterns: painter latches unavailable → same legacy fallback.

## Verification

- **Round 1 (independent adversarial verifier): PASS** — gates green, M1–M4 red, browser acceptance met. Findings fixed in `b947e72e` (watchdog coverage gap: mutation M5 stayed green → dedicated not-parked/no-frames test, now red; mono-stack memo staleness; mid-hold reduced-motion freeze).
- **Round 2 (independent verifier, fresh audit): PASS** — gates green (201 files / 1919 tests at its tip), **M1–M5 all red** with clean isolation, browser re-check met (integration ratio 2.010). Its two P3 notes (per-frame key churn, delimiter-ambiguity) fixed in `8ad4db02` (field comparison + JSON key).
- Gate at the current tip: vitest **201 files / 1921 tests green**, `vue-tsc --noEmit` clean.

## Riding repair (master regression)

`51ee7a85` drops the stray closing brace that #564 left at the tail of `HkTimeline.scss` — the jammed self-hosted CI queue never ran Tests on #564, so the scss-compiling contract suite is red on master and on every branch rebased onto it. One line; this branch does not otherwise touch the timeline.

## Tests

- `HkInput.password.test.tsx`: per-canvas recorder (screenshot-safety invariant, motion across frames, glyphs only offscreen, release stops the pass, input stays `type=password`), parked-bus instant fallback, not-parked/no-frames watchdog fallback (rAF frozen without parking), mono-font cache contract on the new path.
- `revealKinematogram.test.ts`: pure wrap/layout/drift math (surrogate pairs, fit floor 0.4, size band, drift accumulation, random phase seeding, no-2d-context safety).
- `animationBus.test.ts`: `isAnimationParked` guard test.

## Version

`@celestia-island/hikari` → **0.55.9** (master has moved to 0.55.8 and npm 0.55.3–0.55.8 are all published by the sibling in-flight branches; 0.55.9 is the next free slot).

**Merge note:** rebased onto master `fd2d5637`→`5a97461b` (version line re-resolved to **0.55.13** — 0.55.10–0.55.12 shipped from sibling branches in the meantime; the HkTimeline.scss ride-along was absorbed by master #575). Third verification cycle complete: R1 PASS, R2 PASS (its two low findings fixed: background-pattern guard pinned alone by a mirrored asymmetric mock; paint() retry-semantics docblock), R3 terminal PASS with 5/5 mutations surgically red. Final-tip hosted CI green (cargo fmt+clippy+test, web vitest 205 files / 1966 tests + vue-tsc). The self-hosted Tests lane remains jammed by the runner-farm backlog (node-ci-2 offline) — the hosted fallback lane carries the real runs.


langyo:
**Terminal round finding (previous cycle) — fixed in `a3a8780c`.** The tile canvas kept its default 300×150 backing store, clipping the 512×128 noise ImageData: real pattern period 300 (not the documented 512), transparent rows below 128 device px, and wrapDrift's modulus desynced so the field jumped every wrap. Fix = sizing the canvas before `putImageData`; guards = a structural test (an offscreen 512×128 canvas must exist — mutation red) and a browser wrap check: after ~572 device px of drift (1.1 tile wraps), exactly **one** correlation peak in [0,512) at **1.000** (a 300-period field would show two peaks ~300 apart). Screenshot safety was never affected. Per the verification rules the failed terminal round reset the cycle to zero; a fresh 3-round verification runs on the fixed tip.

PR #566, opened 2026-09-16
🐛 Drop the stray brace the hover-highlight cleanup left in HkTimeline.scss.

langyo:
## Summary

- #564's reduced-motion rewrite left one unmatched closing brace; sass compile broke for every consumer on the 0.55.8 package. Removes the brace.

## Checklist

- [x] Code follows the project's style guidelines
- [x] Self-reviewed the diff
- [x] Added or updated tests for the changes (timeline suite green)
- [x] Updated documentation if needed
- [x] No new warnings introduced

PR #569, opened 2026-09-16
🧪 Re-run checks on the merged fix line.

langyo:
## Summary

Re-runs the check suite over the merged timeline hover-highlight line (the #564 content plus master's #565). The #566 PR range went invalid after its force-push; this branch carries the same tree with a linear base. Needed so the brace fix can ride the 0.55.9 package — 0.55.8 shipped a stray brace that breaks sass for consumers.

## Checklist

- [x] Code follows the project's style guidelines
- [x] Self-reviewed the diff
- [x] Added or updated tests for the changes (timeline suite green on this tree)
- [x] Updated documentation if needed


langyo:
Empty re-check PR — carries zero diff against its base; its tree equals master@#565 (59b04b27) and the hover-highlight line it validated is on master (see #559/#575).

PR #576, opened 2026-09-16
✨ Add track, plain and slider variants to the icon button group.

langyo:
## Why

`HkIconButtonGroup` shipped exactly one look — a bordered inset track — with no knob for hosts to tune it. The 2026-09-17 user direction: a frame is not part of a group's meaning; icon arrays are special in that they work both as a tight strip with a sliding highlight or as mutually independent buttons, and consumers must be able to choose.

## What

New `variant` prop, orthogonal to the semantic `mode`:

- **`track`** (default) — the existing bordered strip, byte-identical for current consumers.
- **`plain`** — no outer frame; independent icon buttons, each tinting its own active state.
- **`slider`** — frameless and **gapless** with a sliding thumb pill behind the active item. The thumb's position is the active index × the `--hk-icon-group-item-size` variable (a hair of inset so rounded corners never clip) — no JS measurement, so hosts retuning the size keep the thumb aligned. Single-selection only: `multiple` mode tints per item (one thumb cannot hop between simultaneous selections); `buttons` mode has no selection and renders no thumb.

Version raised to 0.55.13 (publish rides the `v*` tag flow; consuming PR to follow in shittim-chest — the MFA factor picker on the login step-up wants `variant="slider"`).

## Tests

Five new contract tests pin the variants (default stays track; plain keeps semantics; slider thumb renders aria-hidden, carries the index CSS variable, and moves on selection; thumb suppressed for multiple/buttons/no-selection), plus mutation-sensitivity confirmed by the verification rounds.

---

**Rebase note:** rebased onto master (e10f8d32) to resolve the package.json version collision — this change now rides master's **0.55.14** line; the post-merge `v0.55.14` tag publishes everything in one shot. Gates re-verified on the rebased tip: component tests 13/13, vue-tsc clean.


langyo:
**Queue exemption (§8.3.1, ci-ops):** the self-hosted Rust jobs have been QUEUED for ~2.5 hours (farm congestion). The hosted frontend jobs already PASS (web: vitest + vue-tsc; fmt + clippy + test). Per policy, merging on local verification + lint-commits:

- lint-commits / pr-title-check: PASS.
- Full vue package suite 204 files / 1956 tests green locally (--no-file-parallelism); vue-tsc clean.
- 3-round independent verification (private snapshots; mutations M1–M6 incl. both SCSS-contract pins — every pin surgically red).

Sequencing note: the `v0.55.13` tag will be pushed only AFTER this merge so one publish carries master's #563 bump plus this change (registry latest is still 0.55.12).

PR #577, opened 2026-09-16
✨ Give selection grids a width floor and stop the header promising avatar edits.

langyo:
# PR body — hikari `feat/avatar-veil-grid`

Title: `✨ Give selection grids a width floor and stop the header promising avatar edits.`

## Summary

Two related user-facing fixes for the panel family (user report 2026-09-16):

1. **The admin header's avatar no longer shows a camera hover veil.** The
   trigger opens the account *menu* — the avatar edit is the Camera item one
   row down — so the veil promised click-to-change and lied. Markup, hover
   CSS and the icon class are removed together; the menu item keeps its icon.
2. **`HkSelectionGrid` gains `minItemWidth`** (px): adaptive floor mode. The
   column count derives from the available width (`auto-fill` with a
   `minmax(min(floor, 100%), 1fr)` track), so hosts wide enough for one more
   column simply get it — a `56rem` modal lays three ≥240px options per row —
   while narrow hosts step down to two, then one, without horizontal overflow
   at any width (the `min(..., 100%)` guard caps a lone track at the
   container). Fluid grids emit no `data-cols` hook, so the narrow-viewport
   overrides cannot fight the fluid template; the fixed `columns` contract is
   untouched when `minItemWidth` is unset.

First consumer: shittim-chest's change-avatar window (follow-up PR, lands
after 0.55.14 is published and consumed via the lockfile).

## Verification

- `vitest run` (packages/vue): **199 files / 1902 tests passed** (incl. 4 new
  fluid-mode contract tests + the untouched HkAdminHeader suite).
- `vue-tsc --noEmit`: clean.
- The removal is complete: `veil` survives only in comments; no consumer in
  the family references the removed classes.
- Version: master 0.55.13 → **0.55.14**.
- This branch also carries the one-brace repair for `HkTimeline.scss`
  (#564 left it uncompilable on master; #575/#573 landed the same fix
  upstream mid-flight — the branch was rebased onto that result, so the
  commit is dropped from this PR line).

## Notes for release

- `npm Release` workflow_dispatch with `version=0.55.14` after merge.
- 0.55.14 contains **no breaking changes** for existing consumers (new
  optional prop + the veil nobody should depend on; the veil was purely
  decorative markup inside the trigger button).


langyo:
Merge exemption note (§8.3.1): the ten queued jobs are the self-hosted Rust farm lanes; this diff touches no Rust (packages/vue only — two components, their tests, one scss sheet, version bump). The hosted `web: vitest + vue-tsc` check covers the changed surface and is green (and the full suite was run locally: 199 files / 1971 tests). Farm queue congestion is documented in PLAN §2.2.

PR #580, opened 2026-09-17
✨ Make the password reveal readable with boiling noise and add plain and toggle reveal modes.

langyo:
✨ Make the password reveal readable with boiling noise and add plain and toggle reveal modes.

## Why

The hold-to-reveal introduced in #563 (counter-drifting noise kinematogram) achieved its screenshot-safety goal but failed its actual job: **users could not read their own password**. Root causes, all in the noise design:

1. **1 device-px noise grain** — glyph boundaries made of 1px noise smear into static, especially at dpr 1.
2. **Glyphs capped at 18 CSS px**, shrinking to 0.4× (≈7px) for longer passwords — far below comfortable letterform recognition for noise-textured text.
3. **Direction-opponent drift (±40px/s) as the ONLY segmentation cue** — no luminance/density/contrast difference may exist between glyph and background noise (that is the screenshot-safety invariant), and motion-direction segregation is the weakest cue of the set, degrading further whenever the shared animation bus drops frames.

### Noise strategy: boil instead of counter-drift (default, screenshot-safety invariant unchanged)

- The glyph mask stays still; the **background** noise keeps drifting (−22 px/s); the noise sampled through the glyph mask is **re-sampled at a fresh random tile phase every frame** ("boiling"). Glyphs read as a stable flicker silhouette against a calmly sliding surround — a flicker-vs-drift cue that is far stronger than direction-opponent motion and survives low frame rates.
- Single-frame safety is preserved by construction: both regions sample the SAME tile with the SAME distribution, so any screenshot is statistically uniform noise with nothing for OCR to lock onto. The component test pinning "no password `fillText` ever reaches the visible canvas in noise mode" is unchanged and still enforced.
- Noise is randomized per **2×2 device-px block** (`NOISE_GRAIN = 2`) instead of per pixel — coarse grain carries the temporal signal.
- Glyph layout: size band raised from 12–18 to **13–22 CSS px** (62% of box height), **×1.2 letter-spacing** on every advance (packed password characters were unsegmentable), fit-scale floor raised from 0.4 to 0.5.

### New HkInput props (all opt-in; defaults preserve current behavior)

- `revealStrategy: "noise" | "plain"` — `"plain"` is the industry-standard readable reveal (Windows/macOS/1Password/browsers): the password is drawn as ordinary canvas text while revealed. The DOM input stays `type="password"` (the value is never DOM text beyond the input itself). Motion-free, so it is also the reduced-motion-friendly path. Fully screenshot-visible by design — consumers pick per threat model.
- `revealTrigger: "hold" | "toggle"` — toggle = click to show, click again (or the auto-hide timer) to hide. Keyboard Space/Enter follows the same mode with an auto-repeat guard; the aria-label switches between the existing localized `showPassword`/`hidePassword` strings.
- `revealAutoHideMs` (default 8000, `0` disables) — a toggled reveal never lingers forgotten.

### Behavior edges

- Plain reveal repaints on mid-reveal value edits and on canvas resize even with a parked animation bus (reduced motion); the latched jitter fallback repaints too (it used to keep showing the OLD password until release).
- Clearing the value mid-reveal ends the reveal; disabling the field mid-reveal ends it (the eye unmounts — the reveal must not linger, above all for a toggle with the auto-hide timer off); flipping `revealStrategy`/`revealTrigger` mid-reveal reconciles by ending the reveal (the knobs are read once at reveal start).
- The noise watchdogs from #563 (parked bus → static jitter, mid-hold parking, pattern-less engines) are untouched and still pinned.

## Verification

- Gates: `vitest run` 205 files / 1987 tests green; `vue-tsc --noEmit` clean; `celestia-devtools verify-versions` clean (npm 0.55.15); commit-msg lint clean.
- New/updated tests pin: boiling (a cycling `Math.random` mock drives the exact mask translation sequence — a drift regression goes red deterministically), letter-spacing factor, new size band, plain-mode canvas contract (text on the visible canvas, zero noise machinery, no DOM type flip), plain under reduced motion, mid-reveal repaint (plain AND latched jitter fallback), clear-ends-reveal, disable-ends-reveal, prop-flip-ends-reveal, toggle click/keyboard/auto-repeat/auto-hide/disabled-timer, double-start guard (one painter seed per reveal), and aria-label switching.
- 3-round independent verification with mutation evidence:
  - R1 PASS — 6/6 mutations red (boiling→drift, repeat guard, auto-hide, plain repaint, clear-ends-reveal, visible-canvas fillText); findings were 2 doc staleness nits (fixed).
  - R2 PASS — M7/M8/M10 red; M9 (contrast band) honestly green: unverifiable in happy-dom and uniformity-neutral (affects only human readability, covered by real-browser evidence below); M11 (double-start) honestly green → guard pinned by a new test. Single-frame uniformity re-derived from first principles and holds; the shared-tile shifted self-correlation channel is now named in the threat model (theoretical, dominated by the acknowledged video attack, inherent to the old design too).
  - R3 PASS (terminal review) — full-branch consistency audit clean; M12/M13/M14 red on precisely the pinned behaviors (prop-flip watcher, double-start guard, aria-label switching); 4 comment-wording nits fixed in the final commit.
- Real-browser screenshots (headless chromium against the demo app): single frames of the noise reveal are visually uniform noise; consecutive frames differ in ~86% of pixels (vigorous boil); the plain reveal renders crisply readable text; release/toggle-off restores the dot matrix.

## Notes

- Hold + noise remains the default; nothing changes for existing consumers that don't opt in (HkSignInCard keeps noise+hold).
- Toggling keeps focus in the field (pointerdown preventDefault, unchanged); a toggled reveal persists across blur by design but is always bounded by the auto-hide timer.
- Toggle auto-hide does not re-arm on mid-reveal typing — the deadline from the reveal click stands ("hides a forgotten reveal" is the contract).
- Adoption option for consumers: login surfaces on personal devices may prefer `revealStrategy="plain" revealTrigger="toggle"`; high-exposure surfaces keep the defaults.


langyo:
**Correction to my earlier exemption comment** — the root cause is NOT environmental flakiness; I had it wrong. After rebasing onto master the failure reproduces locally, deterministically:

- #578 (`e10f8d32`) swapped the button loading spinner for the inline `.hk-btn-ring` SVG, updated `HkButton.test.tsx`, but **missed the identical `.hk-spinner` assertion in `HkLoadMore.test.tsx:49`** — so master's suite has been genuinely red since `e10f8d32` (that's what the hosted web job has been reporting on every recent run, including the three master runs cited earlier).
- This PR carries the same one-line fix #578 itself used (`hk-spinner` → `hk-btn-ring`, commit `996c630f`). With it, the full suite is locally green: **205 files / 1992 tests passed**, `vue-tsc` clean.
- Also rebased onto current master and raised the version line to **0.55.16** (`80ac93f6`), since #579 already republished 0.55.15 while this branch was in review.

PR #581, opened 2026-09-17
✨ Put the identity header on the menu rows' content edge in HkMenu surfaces.

langyo:
## Problem

The account identity block (`.s-user-header`, rendered in the admin header's
dropdown header slot and every consumer's `HkMenu` header slot) kept a fixed
`var(--space-14)` inline padding in menu surfaces — a value calibrated for a
row spec the unified menu-item grammar (`_menu-item.scss`) replaced. Rows now
carry a 1px transparent border plus a 12px pad (16px on sheets), and the
grammar's row paddings are **px literals** while the block's was rem. Measured
in the chest account menus (which render exactly this grammar):

| surface | root 16 | root 20 |
|---|---|---|
| desktop popout (HkAdminHeader dropdown / HkMenu header slot) | block 14 vs rows 13 → **1px off** | 17.5 vs 13 → **4.5px off** |
| mobile sheet | 14 vs 17 → **3px off the other way** | 17.5 vs 17 → +0.5px |

The stale comment on `.s-user-header` ("aligns with .s-popup-menu-item's own
space-14") described the pre-grammar world.

## Fix

Two scoped rules in `admin-tokens.scss` (the sheet that owns the block) give
it the rows' content edge **in the grammar's own units**:

```scss
.hk-select-popout-host .s-user-header { padding-inline: calc(1px + var(--hk-menu-item-pad-x, 12px)); }
.hk-select-sheet-panel .s-user-header { padding-inline: calc(1px + var(--hk-menu-item-pad-x, 16px)); }
```

- the shared container insets are rem on both sides and cancel, so the
  formulas hold at **any** root font-size (16 / 12.5 / 20 / 24 all measured 0);
- scoped to the two surfaces that host the block: bare surfaces keep the
  block's own 14px, plain `HkSelect` / `HkPopupSelect` panels are untouched,
  and consumers' drawer-footer context overrides still win;
- `--hk-menu-item-pad-x` is declared on the ROWS, so the fallback literals
  (12/16) are what resolve here — the new contract assertions pin them, and
  the sheet comment says they must follow `_menu-item.scss` if a row padding
  ever moves.

## Contract

`_menu-item.contract.test.ts` (the grammar's anti-drift device) gains an
assertion block: both scoped rules must exist with exactly those formulas,
the sheet branch must not regress to the popout's 12px, and the bare-surface
14px value must stay.

## Version

0.55.16 → **0.55.17** (patch). Consumer follow-up: chest #941 shipped these
same two rules as a declared consumer-side interim — once this releases and
chest picks it up, the chest copies and their doctrine carve-out retire.

## Gates

vitest 205 files / 1993 tests green (contract 8/8), `vue-tsc --noEmit` 0,
`vite build` exit 0.


langyo:
Merging with a §6.3.2 note on the queued self-hosted jobs (Run Tests/Coverage queued ~50m, farm backlog): the **required** check (lint-commits) is green, the GitHub-hosted fallback CI ran and passed, and the full gate set was executed locally on `157fe34b` — vitest 205 files/1993 tests, vue-tsc 0, build 0 — plus independent verification (five contract mutations all red + byte-exact restores; a separate confirmation round re-proved a mutation spot-check, the chest-equivalence of the two rules, and all gates). Queued ≠ failure.

PR #582, opened 2026-09-17
✨ Make the sweep readable window the default password reveal.

langyo:
✨ Make the sweep readable window the default password reveal.

## Why

#580 made the reveal *possible* to read (boiling noise, coarse grain, bigger glyphs) but the field feedback was unambiguous: **users still could not read their own password**. Random-dot kinematograms are inherently hard to read — direction/flicker segmentation is a weak cue at field font sizes, and no amount of tuning fixes that for everyone. This change switches the default to a scheme that keeps the noise aesthetic and part of the capture resistance while being **genuinely readable**: a sweeping readable window.

### `revealStrategy: "sweep"` (new default)

- While revealed, the password renders as ordinary high-contrast canvas text **inside a narrow band** (≈6 advances, clamped into the field, never below a readable floor) that sweeps across the row over the boiling-noise field: constant 220 CSS px/s, a 0.45s dwell at the row end so the tail characters get read, then it loops while held.
- Reading is effortless — real text under the window, motion only in the band position. Reduced motion / parked animation bus / pattern-less engines degrade to the **fully readable static plain text** (never frozen noise, never jitter).
- Capture resistance: a single frame — a screenshot — shows only the characters under the band; the rest of the field stays noise. Weaker than "noise"'s full single-frame safety, but a real bar against casual full-password capture.
- Window geometry is pure math (`sweepWindow`, unit-pinned: pace, end dwell, loop phase, degenerate span) derived from the memoized layout; the band width never collapses below the floor even on degenerate canvas sizes.

### Strategy landscape after this change

| strategy | readability | single-frame screenshot leaks |
|---|---|---|
| `sweep` (default) | effortless | only the band (~a few characters) |
| `noise` (opt-in) | hardest | nothing (statistically uniform noise) |
| `plain` (opt-in) | effortless | everything |

### Also

- The 8 legacy noise-strategy tests now pin `revealStrategy: "noise"` explicitly; the default is pinned by a consumer-path test (HkInput's default, the real public surface).
- Reduced-motion sweep degrade is pinned against the legacy jitter (translates stay 0 — plain, not jitter).

## Verification

- Gates: `vitest run` 205 files / 1999 tests green; `vue-tsc --noEmit` clean; `verify-versions` clean (npm 0.55.18); commit lint clean.
- Real-browser screenshots (headless chromium, demo app): the band shows crisply readable characters ("w 0 r d", "r!Pazzw") while the rest of the field stays noise; the band position advances across frames.
- Independent 3-round verification with mutation evidence (R1: M15–M19 red; R2: M20–M23; R3 terminal review) — see PR comments.

## Notes

- `HkSignInCard` (in-repo consumer) intentionally inherits the new sweep default.
- The reveal-strategy default lives on HkInput (the public surface) and is passed down explicitly; HkPasswordSurface carries an identical defensive default for direct instantiation.


langyo:
## Verification trail (4 independent rounds)

- **R1** @ `aaa4ac2f`: full diff review + M15–M19 all RED (clip removal, window motion, plain-not-jitter degrade, default flip via the HkInput consumer path, winW floor). Findings: version bump was a no-op (0.55.17 already taken by #581) → bumped to 0.55.18; duplicated surface default documented as accepted.
- **R2** @ `ca8bac00`: fresh-eyes re-analysis (perf/inter-strategy/test-quality/docs) + M20–M22 RED; M23 honestly GREEN (two-hold reset unpinned) → test added.
- **R3** @ `630eb6fa`: terminal review — M24 RED (tail-reach pin bites, derivation independently verified sound) but M25 GREEN (two-hold test passed vacuously — no frames driven during hold 1). **FAIL → fixed**: the test now drives 3 real bus frames during hold 1 with a fail-loud self-check; stale comments swept.
- **R4 re-verify** @ `7802a11c`: delta review clean; M25 rerun **RED** (`expected -19 to be -37` — unambiguous 18px gap); vitest 2000/2000, vue-tsc clean, tree pristine.

Real-browser evidence: sweep frames show crisply readable characters inside the moving band ("w 0 r d", "r!Pazzw") with the rest of the field staying noise.

PR #590, opened 2026-09-18
✨ Add a segmented OTP input for one-time codes.

langyo:
## What

Adds `HkOtpInput` (exported as `HOtpInput`, matching the library's dominant `H*` public name), the segmented one-cell-per-character code field the 二次验证 step-up screen needs — and any other one-time code: TOTP, SMS, e-mail, invite, recovery.

**Behaviour**
- Typing advances, Backspace steps back through an empty cell and clears the previous one, ←/→ walk, Home/End jump, Delete clears in place (the row never shifts digits between cells).
- Paste anywhere distributes the whole code from the pasted cell forward; spaces, hyphens, dots, underscores, commas and NBSP are dropped, so the `123 456` shape SMS clients hand out just works. A paste that overruns the row drops its tail.
- `modelValue` is the composed code with no separators. `complete` fires on the short → full edge only (editing a full code does not re-fire it), and `autoSubmit` fires `submitOnEnter` on that same edge, so "submit once the 6th digit lands" is one prop.
- Input is filtered to digits, or `[0-9A-Za-z]` with `alphanumeric`. Multi-character input — autofill, IME, a fast keyboard — spreads across the row instead of being truncated, and an emoji is dropped whole (never cut into lone surrogates).
- A caller-fed `modelValue` goes through the same sanitize + filter and is reported back via `update:modelValue` when the field had to change it, so a host holding the raw SMS shape never silently disagrees with what is on screen. A host that was already correct hears nothing.

**Layout**
- The size ramp is a CEILING, not a fixed width: cells shrink together (staying square via `aspect-ratio`) so the row always fits the card it is in — an 8-cell row in a 290px column, six cells in a 340px phone card, both measured in Chromium.
- The glyph follows the shrunken cell through a measured fit the component publishes per row (`--hk-otp-fitted-font`). CSS cannot do this by itself: a container cannot query its own size and an input has no children to query against, so the row measures one cell and the ramp caps the result through a separate `--hk-otp-font-max`. Measured: 25.7px cells → 13px glyph, 29.25px → 14.8px, 56px → 24px.
- Host hooks: `--hk-otp-cell-size`, `--hk-otp-font-size` (pin the glyph outright), `--hk-otp-gap`, `--hk-otp-gap-group`, reachable as an inline `style` on the component.

**Accessibility**
- The cells are one `role="group"` carrying `aria-label` (or the rendered `label`), so a screen reader announces a single field rather than six stray textboxes; each cell's name comes from `cellAriaLabel(index)` (default "Digit N").
- The first cell advertises `autocomplete="one-time-code"` for SMS autofill; mobile raises the digit pad via `inputmode="numeric"`.
- `error` / `hint` / `label` mirror `HkInput`'s shell, with the error announced through `role="alert"` and the hint wired via `aria-describedby`.
- Focus arriving on a filled cell selects its glyph (keyboard navigation, a click); a forward advance deliberately does not, so the digit just typed is never left in a replace-on-keystroke state.

**Notes for review**
- No i18n in the component: user-facing strings arrive as props (`ariaLabel`, `cellAriaLabel`, `label`, `hint`, `error`), because hikari's context carries only `hikari::*` / `plana::*` keys while the consuming app owns the rest.
- Bumps the package to 0.55.24.

## Verification

- Gates: 208 files / 2093 tests green at the head this PR body describes, `vue-tsc` clean, `componentSelfSufficiency` and `sheetCompile` guards included; `verify-versions` clean (npm 0.55.24).
- 53 component tests cover composition, editing, paste, external values, the accessibility contract, and the two regression classes below.
- Chromium (Playwright 1.63) checks on a chest-shaped auth card: every row reports `scrollWidth === clientWidth`, no cell or hint escapes its card at 1280px and at 390px, cells stay square down to the narrowest row, the focused cell keeps its focus border while hovered, a real clipboard paste of a full code lands distributed, and one stray keystroke after a completed code is a no-op.

## Verification rounds

**R1 (initial head) — FAIL, five findings:** the caret exemption did not hold (`focusCell(…, false)` was defeated by the focus handler's own select-on-entry default: pasting `123456` left the tail cell selected and one stray keystroke rewrote it to `123457`); the glyph never followed a shrunken cell (a container-query clamp that could never run — a container cannot query itself — plus two unit bugs, `parseFloat("1.25rem")` read as a 1.25 *pixel* ceiling); squares broke below ~26px cells; a runtime `alphanumeric` flip left letters in the cells; a sanitized caller value never re-emitted.

**R2 (`6179f5bb`) — PASS, no blocker, no major.** Chromium: both caret paths hold (typed six digits + stray keystroke; real `Ctrl+V` + stray keystroke), navigation still selects, click-then-type replaces (16/16); geometry 28/28; 34 mutations, 20 killed. New minors: a stale focus marker, a glyph pin that could not be released, a length probe corruptible by a host `!important` rule.

**R3 (`f5c8b215`) — FAIL, NO-GO.** Two blockers, both fixed on the current head:

1. *A pin removal never re-fitted.* Pinning used to STOP the fit, which turned the glyph size into state only an event this component never receives could refresh — removing a pin changes no box (silent ResizeObserver) and a host-side `style` mutation renders nothing (no `bindRow`). Chromium measured it: a 20px digit clipped inside a 13px cell, **zero** re-fits across the removal, and the row route (which had worked before the stand-down existed) regressed with it. The fit now always publishes `min(ramp ceiling, host pin, 0.55 × cell)` and the pin still wins in the CSS chain, so the value is a pure function of the current DOM and cannot go stale. Re-measured: pinned at mount publishes `11.0px` (glyph 30px, from the pin), releasing leaves the glyph at the already-correct 11px, nothing clipped.
2. *The focus-marker claim was overstated.* Two of the six required mutations — the `activeElement` guard and the marker's consumption — survived all 57 tests, and `applyIncoming`'s emit guard survived in both directions. All four now die on tests that assert the user-visible outcome (a user re-entering the tail of a completed code replaces its digit: `123457`), not a caret range.

R3's independent re-run on the fixed head is recorded in a PR comment before merge.

**Test portability.** Three caret assertions passed locally and failed on the hosted runner: `selectionStart`/`selectionEnd` maintenance after a synthetic input event is not portable across DOM implementations. They assert observable behaviour now (which keystroke the field accepts or replaces, and what the composed code becomes).

## Process note

R2's campaign initially shared this worktree with my own commit, and my `git add`/`git commit` ran inside one of its mutation windows — the commit captured an in-flight mutant (`stayPut` inverted) and was red on push (3/53). Amended within minutes (`--force-with-lease` against the exact bad SHA); the reruns use a private detached worktree. The rule this yields: a mutation campaign and author commits must never share a worktree.

## Reviewer notes

- **What is verified where.** Unit tests cannot witness four classes of behaviour here, so each is covered by Chromium measurements recorded in the round reports: (a) caret ranges after synthetic events — `selectionStart`/`selectionEnd` maintenance is not portable, so the tests assert accept-vs-replace outcomes instead; (b) `:hover` styling (happy-dom applies no pseudo-classes) — the focus border under a real `page.hover()` is measured in the browser; (c) layout (every box measures 0 in happy-dom) — overflow, squareness and the glyph ratio are measured; (d) stylesheet-driven pin precedence and the mutation observer's non-looping behaviour are measured as style-mutation counts over a quiet second.
- **Known environment-only mutation survivors**, named with severity in the R3 report: the `!important` probe hardening (unit-blind, browser-proven load-bearing: header `11.8px` vs mutant `1.0px`), the CSS-chain anchor, and two pin readers that are equivalent by construction because the row's computed value already reflects them.
- The `length` probe is a DOM `<span>`: a host rule that is ALSO `!important` and targets it can still bias the measurement. That is the documented limit of not using a shadow root; the component's own declaration is `!important`, so a plain host rule cannot reach it.
- The row observes its whole ancestor chain (up to `offsetParent`) for `style` changes, because a pin can sit anywhere above it and removing one leaves no trace on the row. The observer's own writes settle in one pass (measured: zero mutations in a quiet second).


langyo:
## Verification rounds R4–R6 (completion of the rounds section above)

**R4 (`1dec8d76`) — FAIL / NO-GO, two items, both fixed by `b140435c`:**
1. The pin-removal blocker was fixed by *deleting the state that caused it*: the fit now publishes unconditionally (`min(ramp ceiling, host pin, 0.55 × cell)`), the pin still wins the cascade (`--hk-otp-font-size` precedes `--hk-otp-fitted-font`), and a `MutationObserver` over the row's ancestor chain re-fits on pin changes no other observer can see. Chromium: ancestor 12px pin at mount → published 12.0px/glyph 12px; release with no box change and no re-render → published 20.0px/glyph 20px. Zero style mutations in quiet seconds, 1–2 across a pin cycle, adversarial 300→40px squeeze 0/0 — the observer cannot loop with the fit's own writes.
2. A stray tracked copy of the component (`packages/vue/HkOtpInput.tsx`) that a broken mutation-restore had swept into an earlier commit was deleted (PR tree back to 5 files).

**R5 (`b140435c`) — FAIL / NO-GO, test-file only:** the new watcher test could not fail (its assertions followed an explicit `fitGlyph` call; the watcher's own call early-returns because happy-dom measures every box as 0). Fixed by stubbing `getBoundingClientRect` so the watcher alone must move the published value — red under "observer never constructed" and "wrong attribute watched", green on the head. The ramp comment that still named `--hk-otp-font-size` corrected to `--hk-otp-font-max`.

**R6 (`c92eb66a`) — PASS / GO.** Independently reproduced: both unit mutants red on exactly the repaired test; pristine head 60/60 targeted, 208 files / 2101 tests, `vue-tsc` clean; browser 26/26 (incl. F1 12px→20px release, G1 row-pin removal, real Ctrl+V + stray keystroke, focus border under real hover); loop 6/6; compliance 5 files / version 0.55.24 once / `HOtpInput` export / `hk-otp*` classes only; CI runs 35315875097 (CI fallback) and 35315878440 (commit lint) both green on this head.

Known limits, disclosed: caret *ranges* and `:hover`/layout are asserted in Chromium rather than unit tests (not portable across DOM implementations); `R03a`/`R03b` (watcher wiring halves) are unit-invisible by construction, witnessed by browser F1; Safari/Firefox and a consuming-webui smoke were not exercised (the chest pickup PR will exercise the latter).

PR #594, opened 2026-09-18
⬆️ Republish the theme-vars selector fix as 0.55.27.

langyo:
Version-only bump. #593's squash merge landed AFTER the v0.55.26 tag was cut (the tag was cut from a stale local origin/master ref), so the published 0.55.26 tarball does not contain the html:root selector fix. 0.55.27 republishes the fixed tree; consumers on ^0.55.26 pick it up by refreshing their lockfile.


langyo:
CI note: the self-hosted Rust gates (clippy/test/fmt) sit queued on the saturated farm. This PR changes ONE field (packages/vue/package.json version) — its tree is otherwise byte-identical to d7f76a97 (#593), whose hosted fmt+clippy+test job passed and whose local gates ran green (vitest 209 files / 2111 tests, vue-tsc clean). Merging with the documented saturation exemption; admin merge used solely to bypass the queued identical-tree gates.

PR #598, opened 2026-09-19
✨ Give icon group options a success status marker.

langyo:
## What

`HkIconButtonGroupOption` gains `marker: "success"`: a small green dot pinned to the icon button's bottom-right corner (theme success token + a surface ring so it stays legible on active and muted tints), exposed to hosts as the `data-marker` attribute for tests. Purely decorative — the accessible meaning stays in the option's label/tooltip, which the host updates when the marker means "this one is done".

Pairs naturally with the existing per-option `disabled`: chest's two-step 2FA verify panel marks the already-proven factor green and stops it from being re-selected.

Default rendering without `marker` is unchanged.

## Verification

- 2 new tests (15 total green): marker dot + `data-marker` render, untouched sibling stays clean, disabled marked option never fires select.
- `vue-tsc --noEmit` clean.


langyo:
⏳ Environmental CI note per §8.3.1: the Rust matrix and the web vitest/vue-tsc gates are green; the only red is the Test Coverage job — cargo-llvm-cov failed to install upstream (infra), and this PR touches TSX/SCSS only (no Rust surface to cover). The required lint-commits is green; local gates green (vitest 15/15 for the component, vue-tsc clean). Merging with the coverage job waived as environmental.

PR #551, opened 2026-09-16
✨ Give the scheme editor's mode strip a track and an editing-scope hint.

langyo:
## Summary

- The scheme editor's 深色/浅色 strip now has an inset track (8% text-tint background, padded, rounded) so the segmented control reads on any dialog surface — in dark mode the pill-on-same-surface chrome was invisible and the strip looked like two bare text links.
- A muted hint under the strip states the scope: these tabs pick which palette you are EDITING; the app-wide light/dark switch lives in the theme popup's color-mode strip. Localized ×8.

## Checklist

- [x] Code follows the project's style guidelines (rustfmt / clippy clean)
- [x] Self-reviewed the diff
- [x] Added or updated tests for the changes (13 editor/dialog tests green)
- [x] Updated documentation if needed (comments)
- [x] No new warnings introduced

PR #599, opened 2026-09-19
🌐 Land the scheme editor mode-tab hint across all eleven locales.

langyo:
## What

`HkColorSchemeEditor.tsx:421` renders `t("hikari::theme.modeTabHint")`, but the key exists in **zero** locales on master — a dangling i18n reference, so the editing-scope hint under the scheme mode tabs always renders the raw fallback. The scss track and the hint element itself landed via #559 (the scheme-editor strip rework); the locale keys were part of #551, which never merged.

This PR lands the key across **all eleven locales**:

- 8 values are byte-identical to #551's branch (`de/en/fr/ja/ko/ru/zh-Hans/zh-Hant`) — extracted from `origin/fix/scheme-editor-mode-strip` and verified string-for-string.
- 3 new translations (`ar/es/pt`) — #551 simply never covered them; `flatKeyParity` requires the full set, so leaving them out would break key-set parity.

## Notes

- Key placement follows each file's existing tail (append after last key); no reordering.
- Version: npm `0.55.30` (0.55.29 is already taken on the registry by the #598 release line; master's manifest reads 0.55.29).
- #551 to be closed as superseded when this merges — forensics in `/mnt/work/hikari-stale-audit/VERDICTS.md`.

## Gates (worktree, shared install)

- `vitest run`: **209 files / 2135 passed / 0 failed** (i18n suite 16/16 included)
- `vue-tsc --noEmit`: **0 errors**


langyo:
Takeover + merge note: self-hosted farm matrices (Format/Clippy/Test/Verify-versions) queued >15 min on the congested farm — merged under the ci-ops §8.3.1 environmental exemption. Local + hosted coverage of the same ground is green: hosted `web: vitest + vue-tsc` and `fmt + clippy + test` both PASS; local full suite 209 files / 2135 tests exit 0 and vue-tsc 0 (data-only change: 11 locale keys + version). Three independent verification rounds passed (R1 /mnt/work/verify-hk599-r1/, R2+R3 /mnt/work/verify-hk599-r2/).

PR #601, opened 2026-09-19
🐛 Pin multi-toast leaves to their pre-patch boxes.

langyo:
## Problem

Right-top toasts squashed during the exit animation whenever several left at once: later-pinned toasts teleported onto the top slot and overlapped the others mid-fade, and a narrow-max-content toast's bar collapsed to a sliver — the overlapped fading texts read as one squashed multi-line blob (up to the 3-line clamp). First reported from wowsp against the published 0.52.x build, which is byte-identical to master for these files.

## Root cause

HkToast / HkBlockingToast pinned the leaving element from LIVE `offset*` reads inside `onBeforeLeave`. During a multi-element removal the first sibling's `*-leave-active` class (`position: absolute`) lands synchronously inside the same patch pass; the shrink-to-fit stack column re-fits immediately, so the later sibling's live reads froze the already-reflowed box (top shifted to the vacated slot, width re-fit to its own max-content). The live `parent.clientWidth` was equally already re-fit, so even a box snapshot without a snapshot-time anchor would pin the right edge wrong.

## Fix

- Apply the repo's established pre-patch snapshot pattern (HkListTransition / HkTabs): the host's `onBeforeUpdate` records every child's box while the DOM is still the pre-patch tree; the before-leave hook pins from that snapshot.
- `LeaveBoxSnapshot` gains an optional snapshot-time `right` so right-anchored hosts don't compute the anchor against the mid-removal containing block. Additive and backward compatible (left-anchored callers unaffected).
- Regression tests in both suites stub layout getters that branch on the first sibling's leave-active class to emulate the browser's synchronous mid-patch reflow; they fail against live-read pinning (verified) and pass with the snapshot. Plus a `box.right` precedence case in `dom.test.ts`.

## Verification

- `vitest run` toast + dom suites: 20/20 pass; regression tests confirmed failing on unfixed master.
- `vue-tsc --noEmit` clean.
- Real-browser frame-by-frame capture (CDP against the toast stack): before — all leaving toasts pinned to `top: 0` and overlapped at the same rect; after — each toast fades out in its own row (tops 0/95/155 preserved), widths and line counts constant throughout the leave.
- Note: `HkCrossfade.test.tsx > "runs the Transition in default (simultaneous) mode…"` fails on clean master in a Windows/happy-dom environment (pre-existing, unrelated to this change).


langyo:
Merge gate note: the two checks that cover this diff are green (web: vitest + vue-tsc; fmt + clippy + test hosted; plus commit/title lint). The remaining 10 pending jobs have sat in the hosted-runner queue for ~10 minutes with 0s execution time — runner starvation, not a code failure (and the diff touches no Rust sources). Merging under the environmental-failure exemption with local verification: targeted vitest suites 20/20, vue-tsc clean, regression tests confirmed failing on unfixed master.

PR #602, opened 2026-09-19
🐛 Give the slider icon strip its missing inter-item gap.

langyo:
## What

The slider variant of `HkIconButtonGroup` set `gap: 0` (a packed row), so neighbouring items visibly touched — reported on the login verify panel where the passkey / TOTP / password tiles read as one merged control.

A visible gap must feed BOTH sides of the illusion:

1. the flex `gap` between items;
2. the thumb pill's `translateX(index * (item-size + gap))` arithmetic.

One custom property (`--hk-icon-group-slider-gap`, 0.375rem default) now drives the two; hosts retune it per context and the thumb position stays pure index math with no JS measurement. The property is declared in-place on the slider block with a warning comment so the `gap: 0` regression cannot come back silently.

## Verification

- 15/15 component tests green; `vue-tsc` clean.
- Visual: 0.375rem separation between every adjacent item, thumb aligned to the active item including across the gap.

Bump to 0.55.30 (0.55.29 shipped without this — the change was lost in a worktree cleanup before the commit).


langyo:
⏳ Note on the pending/red non-required checks: version-consistency red is EXPECTED for a version-bump PR (package.json = 0.55.29... here 0.55.30 vs npm latest — it turns green only after the tag publishes, same as every previous hikari release); the Rust matrix is queued on the congested farm while the load-bearing hosted gates (fmt+clippy+test, web vitest+vue-tsc) and the required lint-commits are green. Local: 15/15 component tests, vue-tsc clean. Merging per §6.3/§8.3.

langyo:
🔁 Rebased onto master (d7f9beae) and rebumped to **0.55.32** (0.55.31 was consumed by the parallel #599/#601/#603 release wave). Version-consistency stays red until the v0.55.32 tag publishes — expected for a version-bump PR, waived per §6.3/§8.3 (required lint-commits green; hosted fmt+clippy+test and web gates green; local 15/15 + vue-tsc clean).

PR #603, opened 2026-09-19
🌐 Land the two dangling i18n keys and pin the t() usage surface.

langyo:
Follow-up to #599 (R2 verification report): master carried exactly two dangling i18n keys, and — more importantly — the incident's root-cause surface (nothing statically audited the `t()` usage side) had no guard at all. This PR lands both keys and pins that surface.

## The two dangling keys and their inline-fallback semantics

Both call sites pass a **second argument to `t()`**, which `useI18n().t` (i18n/context.ts) resolves as `state.messages[key] ?? enFallback[key] ?? fallback ?? key`. So with the key absent from every locale bundle, runtime showed the hard-coded English literal — never the raw key name, in any locale:

- `hikari::emptyState.loading` — `HkEmptyState.tsx:47`, `t("hikari::emptyState.loading", "Loading")`. Screen-reader-only live-region text for the loading spinner.
- `hikari::filePicker.currentPath` — `HkFileBrowserDialog.tsx:447`, `t("hikari::filePicker.currentPath", "Current path")`. `aria-label` of the breadcrumb nav in the file browser pathbar.

(If a call omits the second arg, a missing key displays the key string itself — verified in context.ts; neither of these two does.)

## Locale landing (11/11)

- `filePicker.currentPath` goes into the nested `components` section in its alphabetical slot (`copyPath` < `currentPath` < `cut`), next to its family.
- `emptyState.loading` is appended as a top-level flat key, the post-0.3 appended-at-bottom style (same as #599's `theme.modeTabHint`).
- en values are the **exact inline fallback strings** ("Loading", "Current path"); the other 10 locales are semantic translations aligned with each locale's existing vocabulary (e.g. `filePicker.loading` → "読み込み中", "加载中", "載入中" … minus the ellipsis en does not carry).
- All 11 files verified byte-identical to canonical `json.dumps(indent=2, ensure_ascii=False)` round-trip; 0 key reorderings at any level (R2 §a/b method).

## The audit nail: `packages/vue/src/i18n/usageSurface.test.ts`

Static scan of every consumer source under `packages/vue/src` (446 `.ts`/`.tsx`/`.vue` files, i18n suite itself excluded) for `t("hikari::…")` literals — regex covers **all three quote forms** (double / single / backtick template) **and line breaks after the opening paren** (HkPhoneInput.tsx:194 already uses that split form; a same-line-only regex silently missed 2 keys). Asserts every referenced key resolves to a non-empty string in the loader-merged bundle universe of **every one of the 11 locales** (loader semantics: per-locale union over all `*.json` of flat keys + one nested level).

Zero-hit discipline (self-proof, red-by-construction): floors pinned near today's real volume (≥400 files / ≥320 sites / ≥260 distinct keys — actual: 446/325/270), plus named positive controls including `theme.modeTabHint` @ HkColorSchemeEditor.tsx and a multi-line-form key.

**Dynamic-key whitelist** (interpolated template keys a static scan can never resolve), enforced **bidirectionally** — every family found in source must be listed, and every listed family must still exist in source, so the list cannot rot:
- `hikari::context.` — HkContextRing closed enum, pinned by name in contextKeys.test.ts
- `hikari::statusBar.backend.` — enumerated connection-state suffixes
- `hikari::statusBar.region.` — optional per-locale Intl.DisplayNames overrides (flatKeyParity LOCALE_SPECIFIC_NESTED)
- `hikari::theme.groups.` — registry type: group/slot ids registered by downstream apps; t() resolves via resolveLocalizedText fallback
- `hikari::theme.tokens.` — closed token enum defined in platform.json

Scope boundary: only `hikari::`-prefixed literals — unprefixed keys ("common.time.*", `${i18nKey}.*`) are the host-app namespace (mergeMessages / props.t), pinned by their own suites. There is deliberately **no static-key whitelist**: a static hikari key must exist in all 11 locales; host vocabulary belongs in the unprefixed namespace.

## Red-then-green evidence (isolated copy, byte-identical restore)

1. **Delete `emptyState.loading` from en/components.json** → usageSurface red (`missing or empty in en`, first-referenced-at HkEmptyState.tsx:47) **and** flatKeyParity red (`ar must not carry flat keys unknown to en`).
2. **Typo the t() reference** (`emptyState.loadng` in HkEmptyState.tsx, bundles untouched) → usageSurface red **while flatKeyParity stays green** — the exact #599/M3 gap this test closes; the self-proof control also went red (pinned positive disappeared from the scan).
3. **Prune a whitelist entry** → bidirectional whitelist check red (`families found in source but missing from the whitelist`).
4. All three mutations restored byte-identical (sha256-verified); the working worktree was never mutated.

## Gates

- `pnpm exec vitest run`: **210 files / 2139 tests passed, exit 0** (baseline 209/2135 + 4 new).
- `pnpm exec vue-tsc --noEmit`: 0 errors.
- `npm run lint`: ok. cargo fmt/clippy/test: n/a — zero `.rs`/`.toml` changes (`packages/i18n`'s Rust locales are a separate directory, untouched).
- Version: `packages/vue/package.json` 0.55.30 → **0.55.31** (registry `dist-tags.latest` = 0.55.30, 0.55.31 free — rechecked immediately before push).


langyo:
Takeover + merge note: exec round (/mnt/work/hk-dangling-exec/) + independent verification PASS (/mnt/work/verify-hk603/): keys byte-match the inline fallbacks, audit pin survived five adversarial probes (typo'd key, fake key, narrowed whitelist, off-whitelist dynamic template, double-quoted concat all RED; single-quote form proven in-scan), gates 2139 green + vue-tsc 0. F1 recorded as follow-up (non-blocking): HkStatusBar.tsx:233 constructs `hikari::statusBar.tier.*` away from its t() call — adjacency-based scanning misses that one family (all 4 keys currently present in 11 locales); a later pin should scan construction sites or add named positives. Farm checks queued per policy; hosted equivalents green.

PR #605, opened 2026-09-19
🧪 Make the crossfade wiring contract EOL-agnostic.

langyo:
## Problem

`HkCrossfade.test.tsx > "runs the Transition in default (simultaneous) mode…"` failed on every Windows machine (reported from the wowsp workspace while verifying #601 locally) while CI stayed green.

The wiring-contract test reads the raw `HkCrossfade.tsx` source and pins a multi-line literal (`'<Transition\n            name="hk-crossfade"'`). Windows clones with `core.autocrlf=true` check the file out with CRLF terminators, so the LF literal could never match — an environment-only red, not a component bug.

## Fix

Normalize CRLF/CR to LF once at the read site. On LF checkouts (CI) the normalization is a no-op, so nothing changes there; on Windows checkouts the contract's shape/indentation pin keeps working as intended.

## Verification

- CRLF Windows checkout: HkCrossfade suite 13/13 (previously 12/13), full vue suite green across two full runs.
- LF behavior untouched by construction; the hosted web job covers it.


langyo:
Merge gate note: checks covering this diff are green (web: vitest + vue-tsc hosted — LF checkout, normalization is a no-op there; fmt + clippy + test hosted; commit/title lint). The remaining 10 jobs have queued ~9 minutes with 0s execution time — the same hosted-runner starvation as #601; the diff is a 4-line test-only change with no Rust surface. Merging under the environmental-failure exemption with local verification (CRLF Windows checkout: HkCrossfade 13/13, full vue suite green).

PR #606, opened 2026-09-19
🐛 Pin the toast leave box to its exact fractional size.

langyo:
## Problem

Follow-up to #601, reported from the wowsp workspace: on real machines (fractional layout widths from Windows display scaling / font metrics) a toast whose single-line message sat within a sub-pixel distance of the wrap boundary would push its **last character onto a second line** the moment the exit animation started — every time, exactly one character.

Reproduced in a real browser against the published build: in-flow box `369.40625px` wide, message fitting with `-0.4px` slack; the leave pin wrote `width: 369px` (the `offsetWidth` integer) and the message rewrapped 1 → 2 lines.

## Root cause

#601's pre-leave snapshot took sizes from the `offset*` family, which **rounds to integers**. Whenever the real box rounds down, the pinned box is narrower than what the user was looking at by up to half a pixel — text within that distance of the boundary rewraps.

## Fix

Measure snapshot sizes with `getBoundingClientRect` — exact, and translation-invariant (the only transforms these surfaces apply are enter/move translations, which do not change a box's size). Positions stay on `offsetTop`/`offsetLeft` (the transform-immune layout slot, where the ±0.5px integer error is invisible), and the `right` anchor is now computed against the wrapper's rect width for the same precision.

## Verification

- New regression test: stubs `offsetWidth: 369` against a `369.40625` rect — fails against the integer snapshot, passes with the exact one; existing multi-leave tests updated and green (14/14).
- Real-browser A/B: before — pinned `369px`, message wraps to 2 lines during the leave; after — pinned `369.406px`, line count unchanged.
- `vue-tsc --noEmit` clean; full vue suite green apart from two known parallel-load timing flakes unrelated to this diff (pass in isolation; also green in the hosted web job).


langyo:
Merge gate note: checks covering this diff are green (web: vitest + vue-tsc hosted; fmt + clippy + test hosted; commit/title lint). The remaining 10 jobs have queued ~10 minutes with 0s execution — the same hosted-runner starvation as #601/#605; the diff touches only the two toast hosts and their tests (no Rust surface). Merging under the environmental-failure exemption with local verification: regression test bites on the integer snapshot, 14/14 toast suites, vue-tsc clean, real-browser A/B confirms the sub-pixel wrap is gone.

PR #609, opened 2026-09-20
✨ Add an opt-in expanded-by-default mode for sidebar menu groups.

langyo:
## What

`HkMenu`'s sidebar variant gains an opt-in **`defaultExpandGroups`** boolean. When set, every collapsible group renders **expanded** until the user collapses it; a user collapse sticks (including across `activeKey`/`items` changes). Without the prop nothing changes: the accordion default (only the group holding the active row auto-opens) is preserved, so all existing consumers — drawer account menus, locale pickers, cascade menus — render exactly as before. No-op in the popup variant.

## Why

The console sidebar (shittim-chest `NavSidebar`) groups its nav into themed sections, but the accordion default means at most one section is open at a time and the rest read as hidden entries. Operators expect a sidebar that reads as a table of contents: everything visible, collapsible on demand. The expansion state machine lives in hikari's `HkMenu`, so the capability goes upstream here (§3.3: extend upstream, then consume); the chest consumer PR flips the prop in a follow-up.

## Semantics (pinned by tests)

- **Default (no prop) — additive no-op**: groups without the active row carry no `data-open` marker and their children stay unrendered (existing behavior, now pinned at DOM level).
- **With the prop**: all groups mount expanded (`data-open` + `aria-expanded="true"`), active row or not.
- **First toggle collapses** (open is the default state in this mode); re-expanding works.
- **No auto-resurrection**: navigating into a user-collapsed group keeps it collapsed — the accordion mode's auto-expand watch stays inert.
- **Group independence**: collapsing one group never touches its siblings' default-open state.
- **Nested groups (depth > 1)**: mount expanded too; a parent collapse/expand round-trip restores the inner group still open (its state lives in the key sets).

## Gates

- `vitest run src/components/HkMenu.test.ts` — **38/38** (33 pre-existing + 5 new; commit 2 added the nested-group pin)
- Full `packages/vue` suite — 210 files / **2149 tests, all green**
- `vue-tsc --noEmit` — clean
- Version: `@celestia-island/hikari` 0.55.33 → **0.55.34**

## Verification

- R1 PASS (independent): diff/state-machine review, all gates, mutations M1–M3 red, clean-tree + exact-HEAD checks.
- R1 MINOR (nested groups unpinned) → fixed by commit 2 before R2.
- R2 PASS (independent, from scratch): re-derived everything on `696aeffe`; mutations M1–M4 (incl. test-side redundancy probe) all red as required; popup isolation traced structurally; commit lint clean.

Consumer: shittim-chest `NavSidebar` (admin console sidebar) to follow once this lands and publishes.


langyo:
CI exemption record (§6.3/§8.3): the self-hosted farm jobs did not run on this PR (cancelled on commit 1 during the documented farm saturation window; no farm checks exist on head). The GitHub-hosted fallback ran the full gate set on commit 1 — `fmt + clippy + test (hosted)` ✅ and `web: vitest + vue-tsc (hosted)` ✅ — and the only delta on head (696aeffe) is a test-only commit (+37 lines, one new nested-group pin, no source change), independently verified locally by three verification rounds (R1/R2/R3): targeted 38/38, full packages/vue suite 210 files / 2149 tests green, vue-tsc clean, mutation evidence M1–M4 all red as required, clean tree + exact HEAD reconfirmed after every mutation. The only required check (lint-commits) is green on head.

